PR #5, opened 2023-08-04
Update Dependency Injection Pattern

crickman:
When using the SemanticMemory AppBuilder, the SemanticMemoryConfiguration is easily available...but requiring configuration for the methods creates friction for applications or services aligned with a more general pattern (not using the SemanticMemory AppBuilder).

This change effectively *inverts* the DI pattern so that config is not required to be provide by the caller.  Conditional evaluation occurs within the factory methods (as opposed to coordinating the invocation of factory methods).

Ultimatley, this retains the original pattern and also more easily enables alternate patterns.


dotnet/CoreLib/AppBuilders/HostedHandlersBuilder.cs
```diff
@@ -17,19 +16,19 @@ public static HostApplicationBuilder CreateApplicationBuilder(string[]? args = n
 
         if (Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT")?.ToUpperInvariant() == "DEVELOPMENT")
         {
-            builder.Configuration.AddJsonFile("appsettings.Development.json", optional: true);
+            builder.Configuration.AddJsonFile("appsettings.development.json", optional: true);
```
dluc, line 19: .NET usually has PascalCase values for "Development" and "Production" in the env var ASPNETCORE_ENVIRONMENT and I think this could be an issue in some case sensitive filesystems. I prefer lowercase too, but I'm not sure what's .NET expectation across OSes.

dotnet/CoreLib/AppBuilders/DependencyInjection.cs
```diff
@@ -120,170 +131,176 @@ public static void UseSearchClient(this IServiceCollection services, SemanticMem
...
+                        config.ContentStorage.AzureBlobs.Container,
+                        serviceProvider.GetService<ILogger<AzureBlob>>());
 
             default:
                 throw new NotImplementedException($"Azure Blob auth type '{config.ContentStorage.AzureBlobs.Auth}' not available");
         }
     }
 
-    public static void UseDistributedPipeline(this IServiceCollection services, SemanticMemoryConfig config)
+    private static QueueClientFactory GetQueueFactory(this IServiceProvider serviceProvider)
```
dluc, line 250: not sure about this. I would ask to ignore the dependency injection approach because I have a lot going on here and planning to make some heavy changes during August.

Is the existing approach blocking the integration with CC?

dotnet/CoreLib/AppBuilders/DependencyInjection.cs
```diff
@@ -120,170 +131,176 @@ public static void UseSearchClient(this IServiceCollection services, SemanticMem
...
+                    default:
+                        throw new NotImplementedException($"Orchestration type '{config.Orchestration}' not available");
+                }
+            });
+
+        return services;
     }
 
-    public static void UseAzureBlobStorage(this IServiceCollection services, SemanticMemoryConfig config)
+    private static IContentStorage UseAzureBlobStorage(this IServiceProvider serviceProvider)
```
dluc, line 222: ditto, see comment below

dotnet/CoreLib/AppBuilders/DependencyInjection.cs
```diff
@@ -120,170 +131,176 @@ public static void UseSearchClient(this IServiceCollection services, SemanticMem
...
 
-        services.AddTransient<SearchClient>(_ => new SearchClient(vectorDb, embeddingGenerator, kernel));
+                return new SearchClient(vectorDb, embeddingGenerator, kernel);
+            });
+
+        return services;
     }
 
-    public static void UseContentStorage(this IServiceCollection services, SemanticMemoryConfig config)
+    public static IServiceCollection UseContentStorage(this IServiceCollection services)
```
dluc, line 158: ditto, see below

dotnet/CoreLib/AppBuilders/DependencyInjection.cs
```diff
@@ -49,68 +49,79 @@ public static SemanticMemoryConfig UseConfiguration(this IServiceCollection serv
...
+            services
+                .UseContentStorage()
+                .UseOrchestrator()
+                .UseKernel();
+    }
 
-        ISemanticMemoryVectorDb vectorDb;
-        ITextEmbeddingGeneration embeddingGenerator;
-        IKernel kernel;
+    public static IServiceCollection UseKernel(this IServiceCollection services)
```
dluc, line 61: ditto, see below

dotnet/CoreLib/AppBuilders/DependencyInjection.cs
dluc, line 1: Could we leave these changes out of the PR? see comments below

dluc (changes requested):

dluc:
@crickman I agree the current DI approach is not optimal, and I've been planning to clean it up soon. If we merge this PR I cannot promise I will still make some heavy changes. So ultimately, unless the integration is broken, I would ask to not make changes to `DependencyInjection.cs` for now.

dotnet/CoreLib/AppBuilders/DependencyInjection.cs
crickman: Yes, it does create a bit a friction with CC.  To me, the existing DI seems a little upside down.  Honestly, I think this sets the stage for better evolution.

For more complex injection patterns, introducing provider pattern can often simplify how this is managed...as opposed to creating more injection points.

dotnet/CoreLib/AppBuilders/HostedHandlersBuilder.cs
crickman: Yeah, that occurred to me as an afterthought also...I'll rollback.

crickman:
> @crickman I agree the current DI approach is not optimal, and I've been planning to clean it up soon. If we merge this PR I cannot promise I will still make some heavy changes. So ultimately, unless the integration is broken, I would ask to not make changes to `DependencyInjection.cs` for now.

Yes, I experienced steep friction on integration...this is what triggered me to look at this.  Honestly, I think this sets you up for a better path for future evolution.  As we discussed yesterday, I do expect to absorb breaking changes from SM over the next few weeks...so no problem there.

I'll follow-up offline also.

dotnet/CoreLib/AppBuilders/DependencyInjection.cs
crickman: Everything I touched is linked/bound according to the original implementation

dluc (approved):

crickman:
@dluc - Merged to current

dluc (approved):

PR #12, opened 2023-08-09
Expose memories

crickman:
SearchClient manages prompt dynamics.  This is incompatible with Copilot Chat which has its own prompt management...as many other applications might.  Copilot Chat requires raw access to memories with citation.

Extracted MemoryClient from existing SearchClient (more S.O.L.I.D).

image


dluc (changes requested):
as per offline discussion, this is no needed. Citations are already sorted. CC should use the SemanticMemoryService class, avoiding the internal SearchClient. The goal of the service is producing an answer, not just a list of citations to process again on CC side.

dluc:
`ISemanticMemoryClient.SearchAsync` now provides the list of relevant sources, across all the memory clients (`Memory`, `MemoryService` and `MemoryWebClient`). That should provide the list of memories.

PR #19, opened 2023-08-15
🔧 Fix directory prompt in InteractiveSetup

xbotter:
 The directory prompt in InteractiveSetup was fixed to correctly ask for the directory where files should be stored.


dluc (approved):
thanks! ☺️

PR #30, opened 2023-08-26
Virtual Memory - Initial Draft

crickman:
Enable a local-only option for development  

Port a basic volatile memory option based on SK framework


dluc:
I pushed to main the simple vector DB I wrote last year (see `Microsoft.SemanticMemory.MemoryStorage.DevTools.SimpleVectorDb`), which allows to store vectors on disk, without running any service. 

If you need a version that stores data in volatile memory, I would add an option to `SimpleVectorDbConfig` and implement a version of `ISimpleStorage` that holds data in RAM. This way you won't have to write another implementation of `ISemanticMemoryVectorDb`, just provide a different underlying data store for SimpleVectorDb.

PR #34, opened 2023-08-31
Update filter type to ICollection<MemoryFilter> to support grouped logical operations

anthonypuppo:

Update vector DB search methods to pass an `ICollection<MemoryFilter>`. Each `MemoryFilter` passed will have its filters AND'd together, while these grouped operations will be OR'd. This modification provides basic support for logical operations, moving away from the previous limitation of hardcoded AND/OR filter criteria.

- New `filters` argument in type `ISemanticMemoryVectorDb` to allow passing `ICollection<MemoryFilter>`.
- `AzureCognitiveSearchMemory` is updated accordingly to accommodate this change.
- It's necessary to update `QdrantMemory` as well. However, as I lack experience with QDrant, I would appreciate if someone could assist with updating that one (or both).
- Ideally, an overload for a single method filter should exist. However, given that the filter argument is already optional, creating an overload with just the filter type changed could lead to ambiguous method issues. This could be a topic for further discussion.


anthonypuppo:
@dluc pinging as this relates to our Discord conversation.

dluc:
looks good, I think we need a functional test to verify that filters work as intended. Could you create one? Here https://github.com/microsoft/semantic-memory/blob/main/tests/FunctionalTests/VectorDbComparison/TestCosineSimilarity.cs you can find some code to get started. I would test the AND/OR combinations for ACS, then we can do Qdrant and SimpleVDB separately.

dotnet/CoreLib/MemoryStorage/AzureCognitiveSearch/AzureCognitiveSearchMemory.cs
```diff
@@ -571,5 +565,24 @@ private static double SimilarityToDistance(double similarity)
...
+                .Select((keyValue) =>
+                {
+                    var fieldValue = keyValue.Value?.Replace("'", "''", StringComparison.Ordinal);
+                    return $"tags/any(s: s eq '{keyValue.Key}{Constants.ReservedEqualsSymbol}{fieldValue}')";
+                });
+
+            conditions.Add($"({string.Join(" or ", filterConditions)})");
+        }
+
+        return string.Join(" and ", conditions);
```
anthonypuppo, line 584: Should this AND/OR logic be reversed? I feel like the more common use case would be to create a single MemoryFilter from search criteria (ie. `QueryHelpers.ParseQuery(uri.Query)`) which in most cases would expect an AND. As it exists now, OR'ing criteria within the same filter seems counter intuitive.

anthonypuppo: This is actually the case in the web client service so I'm thinking it definitely does need to get switched.

anthonypuppo:
- TODO
    - Update examples
    - Tests
- It's not very clean having to create a new list object everywhere for single filter scenarios (`new List<MemoryFilter>() { filter }`) but I don't have a solution for providing an overload without causing ambiguous method errors. Possibly another helper method within the MemoryFilters class?
- MemoryWebClient/QDrant currently just throw exceptions if more than a single filter is passed. To me this makes sense in the context of MemoryWebClient (at least in the interim) but QDrant could definitely be updated by someone who is more familiar with the engine.

dotnet/CoreLib/MemoryStorage/AzureCognitiveSearch/AzureCognitiveSearchMemory.cs
dluc: All rules within one filter are "must have", e.g. use `AND`

dluc:
> * TODO
>       
>       * Update examples
>       * Tests
> 
>     * It's not very clean having to create a new list object everywhere for single filter scenarios (`new List<MemoryFilter>() { filter }`) but I don't have a solution for providing an overload without causing ambiguous method errors. Possibly another helper method within the MemoryFilters class?
> 
>     * MemoryWebClient/QDrant currently just throw exceptions if more than a single filter is passed. To me this makes sense in the context of MemoryWebClient (at least in the interim) but QDrant could definitely be updated by someone who is more familiar with the engine.

I'll update the Qdrant client, no worries.

The new filtering logic should be implemented across all the clients using the memory interface, which should be:

* the memory service web api supports incoming requests with multiple filters
* the memory web client supports sending multiple filters (this should be as easy, JSON serialization does most of the work)
* the serverless client

I agree that this could make the C# API uglier, and I was thinking we could either overload or add a helper. I think overloading is better for discoverability. You can also leave that out of the PR and we'll improve the API later.

There are also a bunch of examples in the repo which needs to be updated, in particular examples `001` and `002` should remain functional.

dotnet/Service/Program.cs
```diff
@@ -132,7 +133,7 @@
                 ILogger<Program> log) =>
             {
                 log.LogTrace("New search request");
-                MemoryAnswer answer = await service.AskAsync(question: query.Question, index: query.Index, query.Filter);
+                MemoryAnswer answer = await service.AskAsync(question: query.Question, index: query.Index, filters: new List<MemoryFilter>() { query.Filter });
```
dluc, line 136: pls update the `MemoryQuery` class to support the same list of filters

dotnet/CoreLib/Search/SearchClient.cs
```diff
@@ -143,7 +143,12 @@ await foreach ((MemoryRecord memory, double relevance) in matches.WithCancellati
         return result;
     }
 
-    public async Task<MemoryAnswer> AskAsync(string index, string question, MemoryFilter? filter = null, CancellationToken cancellationToken = default)
+    public async Task<MemoryAnswer> AskAsync(
+        string index,
+        string question,
+        double minRelevanceScore = 0,
```
dluc, line 149: I think we need a different approach to support the min score and a list of filters.
something like

```
class MemoryFilter :  : TagCollection {
   ..like MemoryFilter now, without the relevance value...
}

// new class, not sure about the name
class MultiMemoryFilter {

    float MinRelevance

    IList<MemoryFilter> Filters

    // we could move "limit" here too, but perhaps on a different PR

}
```

dotnet/CoreLib/MemoryStorage/AzureCognitiveSearch/AzureCognitiveSearchMemory.cs
anthonypuppo: Changed.

dotnet/Service/Program.cs
anthonypuppo: Done.

anthonypuppo:
Note, modified this to use `ICollection<MemoryFilter>` over `IList<MemoryFilter>` as index based lookups are unnecessary.

dotnet/CoreLib/Search/SearchClient.cs
anthonypuppo: Should this be refactored in a separate PR? Feel like the method overloads, configuration, etc could all get revamped as a dedicated change.

anthonypuppo:
@dluc This should be complete minus QDrant support (current implementation throws if more than a single filter is passed). I'd argue some of the cleanup (option refactoring, method overloading, test logic consolidation, etc.) should be separate changes from this but I can implement if you think it's necessary.

dluc:
@anthonypuppo just a quick heads up, I'm out of the office and I will pick this up asap in a couple of weeks 👍

anthonypuppo:
> @anthonypuppo just a quick heads up, I'm out of the office and I will pick this up asap in a couple of weeks 👍

No rush at all, thanks!

dotnet/ClientLib/MemoryWebClient.cs
```diff
@@ -191,26 +192,29 @@ public async Task DeleteDocumentAsync(string documentId, string? index = null, C
...
+        if (filter != null)
+        {
+            if (filters == null) { filters = new List<MemoryFilter>(); }
+
+            filters.Add(filter);
+        }
+
+        if (filters is { Count: > 1 })
+        {
+            throw new ArgumentException("The method does not support a filter list greater than 1", nameof(filters));
```
dluc, line 213: any reason for this exception? what's missing in order to support multiple filters?

anthonypuppo: Think this is outdated per my first pass on this. Will review and modify appropriately in a bit.

anthonypuppo: Removed as there shouldn't be a limitation here. Think I absent mindedly added as a fail-fast thinking the underlying provider could have been QDrant.

dluc: same in `AskAsync` right? 

I'll take care of that one because it's breaking some functional tests

anthonypuppo:
> @anthonypuppo
> 
> > * TODO
> >   
> >   * Update examples
> >   * Tests
> > * It's not very clean having to create a new list object everywhere for single filter scenarios (`new List<MemoryFilter>() { filter }`) but I don't have a solution for providing an overload without causing ambiguous method errors. Possibly another helper method within the MemoryFilters class?
> > * MemoryWebClient/QDrant currently just throw exceptions if more than a single filter is passed. To me this makes sense in the context of MemoryWebClient (at least in the interim) but QDrant could definitely be updated by someone who is more familiar with the engine.
> 
> * Examples updated
> * Functional tests added, both Serverless and Service via WebClient
> * Re-added "filter" parameter (additional to "filters") so the syntax for a single filter is not too verbose
> * MemoryWebClient fixed
> * Qdrant: will do in a separate PR

Looks good to me! Thanks for your help with this.

dluc (approved):

PR #35, opened 2023-08-31
Expose app from MemoryClientBuilder

TaoChenOSU:
When deploying the service to web service which runs in IIS, we believe that we are experiencing the same issue detailed in this post: https://techcommunity.microsoft.com/t5/iis-support-blog/asp-net-core-503-server-has-been-shutdown/ba-p/3830338

The MemoryClientBuilder builds a separate app or builds the app builder that's passed in. We expose the app property so that we can avoid having two apps.

### DIFF
--- a/dotnet/Service/Program.cs
+++ b/dotnet/Service/Program.cs
@@ -38,11 +38,12 @@ appBuilder.Services.AddSwaggerGen();
 
 // Inject memory client and its dependencies
 // Note: pass the current service collection to the builder, in order to start the pipeline handlers
-ISemanticMemoryClient memory = new MemoryClientBuilder(appBuilder.Services).FromAppSettings().Build();
+var memoryClientBuilder = new MemoryClientBuilder(appBuilder).FromAppSettings();
+var memory = memoryClientBuilder.Build();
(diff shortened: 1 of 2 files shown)


dotnet/Service/Program.cs
```diff
@@ -38,11 +38,12 @@
 
 // Inject memory client and its dependencies
 // Note: pass the current service collection to the builder, in order to start the pipeline handlers
-ISemanticMemoryClient memory = new MemoryClientBuilder(appBuilder.Services).FromAppSettings().Build();
+var memoryClientBuilder = new MemoryClientBuilder(appBuilder).FromAppSettings();
```
crickman, line 41: Has this service been deployed outside of dev-box? (I think it'd also be susceptible to the 503 error.)

dluc:
As per offline conversations, the cause of the crashes is Azure VMs with Windows/IIS unable to cope with the load of async queues (threads, number of connections). For instance, using the same VM size with Linux the code runs just fine, and increasing VM size (and cost) with Windows the service survives longer, but eventually crashes after a few minutes. If by any chance there's a memory leak it seems to be affecting only Windows deployments.

PR #40, opened 2023-09-01
Clarify this is a .NET library

brunoborges:
Semantic Kernel has implementations in Python, Java, and TypeScript (TS is WIP), along with .NET. This PR clarifies that SM is targeted only at SK for .NET

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Semantic Memory
 
-**Semantic Memory** is an open-source library and [service](dotnet/Service/README.md)
+**Semantic Memory** is an open-source .NET/C# library and [service](dotnet/Service/README.md)
 specialized in the efficient indexing of datasets through custom continuous data
 pipelines.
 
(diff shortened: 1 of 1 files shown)


dluc:
@brunoborges we don't want to position this as a .NET solution, rather as a service that can be consumed from any language as long as one can use HTTP.  
The serverless client will be C# only but it's an edge case that eventually won't provide all the features possible by hybrid pipelines. Soon the service will leverage pipeline handlers written in any language, and the recommended use case will be using the Memory Web client (the repo will contain a version for a few languages) or as a Plugin for Copilot chats and ChatGPT.

brunoborges:
In this case, Semantic Memory should prrgaps be positioned similarly to how Apache Kafka and many other solutions are: as a tool one can run, and integrate by using clients (in many languages).

While Kafka, as an example, is implemented in Java, it is clear that that is just an implementation detail when one looks at the Kafka website.

In here, on the other hand, it is positioned as a .NET Service, which is why it is confusing.

dluc:
I made a quick update to the README. Please bear in mind this is heavily under development and a few pieces are still in the works. Open to continue the conversation offline or over an Issue, I'll close the PR for now. Thanks!

PR #43, opened 2023-09-03
Leverage NuGet central package management

anthonypuppo:
Refactor to use NuGet central package management ([reference](https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management)). This will simplify dependency management as the project evolves.

- Add `Directory.Packages.props`
- Add `nuget.config` (to adhere to [best practices](https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file))
- Remove explicit package versions from project files


crickman:
Thank you for this wonderful hygiene.  @dluc / @alliscode, I  up vote this for merging.

crickman (approved):

TaoChenOSU (dismissed):

examples/101-using-core-nuget/101-using-core-nuget.csproj
```diff
@@ -8,7 +8,7 @@
     </PropertyGroup>
 
     <ItemGroup>
-        <PackageReference Include="Microsoft.SemanticMemory.Core" Version="0.0.230827.2-preview" />
```
dluc, line 11: in this case please leave the version here, so we can test the nuget on this project without affecting other places.

Directory.Packages.props
```diff
@@ -0,0 +1,75 @@
...
+    <PackageVersion Include="Microsoft.Extensions.Configuration.Binder" Version="7.0.4" />
+    <PackageVersion Include="Microsoft.Extensions.Configuration.Json" Version="7.0.0" />
+    <PackageVersion Include="Microsoft.Extensions.Configuration.EnvironmentVariables" Version="7.0.0" />
+    <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="7.0.0" />
+    <PackageVersion Include="Microsoft.Extensions.Hosting" Version="7.0.1" />
+    <PackageVersion Include="Microsoft.Extensions.Logging" Version="7.0.0" />
+    <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="7.0.1" />
+    <PackageVersion Include="Microsoft.Extensions.Logging.TraceSource" Version="7.0.0" />
+    <PackageVersion Include="Microsoft.SemanticKernel" Version="0.23.230906.2-preview" />
+    <PackageVersion Include="Microsoft.SemanticMemory.Core" Version="0.2.230901.2-preview" />
```
dluc, line 24: I would avoid this row, it will be confusing having both a link to project and to the nuget

Directory.Packages.props
```diff
@@ -0,0 +1,75 @@
+<Project>
+  <PropertyGroup>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+  </PropertyGroup>
+  <ItemGroup>
```
dluc, line 5: could you separate libs from analyzers in two groups please?

dluc (changes requested):

Directory.Packages.props
anthonypuppo: Updated (used SK as a reference).

Directory.Packages.props
anthonypuppo: Not sure I understand this. Package and project references use two different nodes (`PackageReference` vs `ProjectReference`). Since the Semantic Memory package is being referenced in at least once place I feel it makes sense to have at least one version defined centrally in case future projects need to use it. It can still be overridden on a per project basis or ignored completely (and the local project used) by using `ProjectReference`.

examples/101-using-core-nuget/101-using-core-nuget.csproj
anthonypuppo: Do you mean explicitly define the NuGet package version? Wouldn't this defeat the purpose of centrally managing as it wouldn't take advantage of any changes done to the centrally configured package?

Directory.Packages.props
dluc: We use the nuget only in one test project, as an example, and for now we don't have plans about referencing the nuget from within the repo (aside from the example). We might need examples to use different versions for other future reasons, so I would prefer leaving this out. I find it also confusing and it could lead to unexpected headaches debugging an unexpected dependency on the package rather than the code. Using this nuget is very much an exception, and it's better not having it here.

examples/101-using-core-nuget/101-using-core-nuget.csproj
dluc: FWIW I often work on this project on its own, isolated from settings outside the folder, and I find it easier to have all the settings locally. Please leave this entry here, also see other comment about not having the entry in directory.packages.props. Thanks!

Directory.Packages.props
anthonypuppo: Removed.

anthonypuppo:
> I fixed the conflicts, but there is some more work to do, to address errors and warnings:
> 
> * CoreLib.csproj : error NU1008: Projects that use central package version management should not define the version on the PackageReference items but on the PackageVersion items: Microsoft.SemanticKernel. [SemanticMemory.sln]
> * examples/101-using-core-nuget/101-using-core-nuget.csproj : warning NU1604: Project dependency Microsoft.CodeAnalysis.Analyzers does not contain an inclusive lower bound. Include a lower bound in the dependency version to ensure consistent restore results. [SemanticMemory.sln]
> * examples/101-using-core-nuget/101-using-core-nuget.csproj : warning NU1604: Project dependency Microsoft.CodeAnalysis.CSharp.CodeStyle does not contain an inclusive lower bound. Include a lower bound in the dependency version to ensure consistent restore results. [SemanticMemory.sln]
> * examples/101-using-core-nuget/101-using-core-nuget.csproj : warning NU1604: Project dependency Microsoft.CodeAnalysis.NetAnalyzers does not contain an inclusive lower bound. Include a lower bound in the dependency version to ensure consistent restore results. [SemanticMemory.sln]
> * examples/101-using-core-nuget/101-using-core-nuget.csproj : warning NU1604: Project dependency Microsoft.VisualStudio.Threading.Analyzers does not contain an inclusive lower bound. Include a lower bound in the dependency version to ensure consistent restore results. [SemanticMemory.sln]
> * examples/101-using-core-nuget/101-using-core-nuget.csproj : warning NU1604: Project dependency Roslynator.CodeAnalysis.Analyzers does not contain an inclusive lower bound. Include a lower bound in the dependency version to ensure consistent restore results. [SemanticMemory.sln]
> * examples/101-using-core-nuget/101-using-core-nuget.csproj : warning NU1604: Project dependency Roslynator.Formatting.Analyzers does not contain an inclusive lower bound. Include a lower bound in the dependency version to ensure consistent restore results. [SemanticMemory.sln]

This was caused by opting `101-using-core-nuget` out of CPM. Can't tell if it's a NuGet bug or not, but it seems as if for whatever reason it's still trying to pull in the analyzer packages. Possibly because of the `PackageReference` elements in `Directory.Packages.props`? NuGet may be getting confused and not honoring the CPM opt out and still try to install what it thinks are [global](https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management#global-package-references) dependencies.

dluc:
Warnings should be fixed, see last push. I think this is now ok to merge

dluc (approved):

TaoChenOSU (approved):

PR #53, opened 2023-09-13
Add TextPartitioningOption  to TextPartitioningHandler

### LINKED ISSUE #52: Customize chunk size in TextPartitioningHandler
I hope to be able to reduce the size of the chunk to control token consumption during Ask.

xbotter:
Changes Made:
- Added TextPartitioningOption class to represent options for text partitioning.
- Added WithOption method to MemoryClientBuilderExtensions to allow users to customize options.
- Updated the TextPartitioningHandler constructor to accept an optional TextPartitioningOption parameter.
- Updated the TextPartitioningHandler to use the custom text partitioning options when partitioning text files.

closes #52


crickman:
Funny, I was looking at this exact thing last night also...thank you for providing this change.

crickman (approved):

dluc (changes requested):
What about a different approach, allowing to set these value when sending files/data? Allowing to configure these values at start time is a step forward from hardcoded values, but they are still hard coded in a sense. One goal of the service is allowing to experiment with different strategies (for chunking and other tasks), and it could be achieved in multiple ways:
* multiple handlers, very similar in code, only with different settings - e.g. one different handler for each choice of partition size (100, 200, 300, 400, etc) - I think this would become easily too expensive to maintain, with lot of code duplication
* multiple handlers, very different in code, accepting request settings, e.g. the partition size - e.g. a basic partition handler using size, another handler using AI to partition data, etc.

I think we should go in a direction where only the infrastructure is harcoded (either as code or config) and where data processing is configurable on the fly, without restarting code or adding handlers. See for example how one can choose which handlers to use.

For now, if you need different partitioning settings, the best option is creating a custom handler with a different name and different values. With some work, we need to allow the same with a single handler, passing values during the upload (including http requests).

crickman:
Are we sure we don't want to fix hard-coded configuration parameters as the _easiest_ path?

dotnet/CoreLib/Handlers/TextPartitioningOption.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.SemanticMemory.Handlers;
+
+/// <summary>
+/// Represents options for text partitioning.
+/// </summary>
+public class TextPartitioningOption
```
glahaye, line 8: Maybe rename this to TextPartitioningOptions since there are more than a single option

examples/105-dotnet-CustomOptions/Program.cs
```diff
@@ -0,0 +1,21 @@
...
+    .WithOption(new TextPartitioningOption()
+    {
+        MaxTokensPerLine = 100,
+        MaxTokensPerParagraph = 300,
+        OverlappingTokens = 50,
+    })
+    .BuildServerlessClient();
+
+await memory.ImportDocumentAsync(new Document()
+    .AddFile("mswordfile.docx"), steps: new string[]
```
glahaye, line 17: Please use a smaller docx file.  We don't need a Word document with pictures for this test. If we keep this file, we'll increase the size of our repo by more than half a meg - even if we end up deleting this file in the future!

dluc: this is actually a copy of a file already in the repo which we forgot to compress :-) I'll take care of it

dotnet/CoreLib/AI/OpenAI/DependencyInjection.cs
```diff
@@ -47,6 +48,12 @@ public static MemoryClientBuilder WithOpenAITextGeneration(this MemoryClientBuil
         builder.Services.AddOpenAITextEmbeddingGeneration(config);
         return builder;
     }
+
+    public static MemoryClientBuilder WithOption<T>(this MemoryClientBuilder builder, T option) where T : class, new()
```
dluc, line 52: this method should probably be in the builder and not under AI/OpenAI

dluc: moved

dotnet/CoreLib/Handlers/TextPartitioningOption.cs
dluc: fixed

dluc (approved):

PR #63, opened 2023-09-19
Flag messages being processed in SimpleQueues

TaoChenOSU:
Address issue https://github.com/microsoft/semantic-memory/issues/61

### DIFF
--- a/dotnet/CoreLib/Pipeline/Queue/DevTools/SimpleQueues.cs
+++ b/dotnet/CoreLib/Pipeline/Queue/DevTools/SimpleQueues.cs
@@ -7,6 +7,7 @@ using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Timers;
+using DocumentFormat.OpenXml.Presentation;
 using Microsoft.Extensions.Logging;
 using Microsoft.SemanticMemory.Diagnostics;
 using Timer = System.Timers.Timer;
@@ -38,6 +39,9 @@ public sealed class SimpleQueues : IQueue
 
     // Sorted list of messages (the key is the file path)
     private readonly SortedSet<string> _messages = new();
+
+    // List of messages currently being processed (the key is the file path)
+    private readonly HashSet<string> _processingMessages = new();
 
     // Lock helpers
     private readonly object _lock = new();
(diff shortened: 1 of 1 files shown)


alliscode (approved):

dotnet/CoreLib/Pipeline/Queue/DevTools/SimpleQueues.cs
```diff
@@ -177,16 +186,22 @@ private void PopulateQueue(object? sender, ElapsedEventArgs elapsedEventArgs)
                 FileInfo[] files = d.GetFiles($"*{FileExt}");
                 foreach (FileInfo f in files)
                 {
-                    this._log.LogInformation("found file {0}", f.FullName);
-                    this._messages.Add(f.FullName);
+                    if (!this._processingMessages.Contains(f.FullName))
```
dluc, line 189: @TaoChenOSU why this conditional? `_messages` should already contain a message if it's being processed, and even if we add it again, it shouldn't cause duplicate processing (or errors)

TaoChenOSU: You're right! `_messages` contains all the messages without duplication since it's a set. The problem arises from `DispatchMessages` where it doesn't check if a message is being processed, and this will cause errors as described in https://github.com/microsoft/semantic-memory/issues/61. 

And we have no way to mark if a message is being processed. That's why I added the `processingMessages` set.

The condition here is mostly for the logging statement within this condition. While a message is being processed, it will keep populating the same message to the set. Although the set will not change, but it will produce many logging information.

dluc: if the check is only to avoid logging, we could check _messages instead of _processingMessages, right?

```
if (!this._messages.Contains(f.FullName))
{
    this._log.LogTrace("Found file {0}", f.FullName);
    this._messages.Add(f.FullName);
}
```

TaoChenOSU: Yes, this is correct. In fact, I think it makes more sense to use `_messages`.

PR #66, opened 2023-09-26
Change queues content in async design

dluc:

In the current design, the source of truth for async pipeline **status** is stored in the queues as a message moving from step to step. Since the initial design, the size of the status has been growing considerably, e.g. containing extra information about each partition, how to update docs, etc. The size of this object can now reach several KBs and cause errors with queueing services designed to work with small payloads. In particular errors are already happening when importing long documents. One reason for the current design was to increase the resilience of the service, not to depend on the state on disk.

There are a couple of approaches to fix the problem:
1. reduce the size of the status object, using an additional file, which would need to be read every time
2. remove all the information from the message in the queue, passing only a pointer (index name + pipeline ID) used to load the status from disk.

The PR contains approach #2 with the relevant changes, considering that the source of truth is now the status information stored on disk (which previously was sort-of ignored/overwritten).

The PR includes also some new functional tests to easily check the service.


dluc:
This should address and fix https://github.com/microsoft/semantic-memory/issues/57

dotnet/CoreLib/Pipeline/DistributedPipelineOrchestrator.cs
```diff
@@ -167,28 +184,30 @@ public override async Task RunPipelineAsync(DataPipeline pipeline, CancellationT
...
+            // to reduce the chance of the persisted state to be out of sync.
             using IQueue queue = this._queueClientFactory.Build();
             await queue.ConnectToQueueAsync(nextStepName, QueueOptions.PublishOnly, cancellationToken).ConfigureAwait(false);
-            await queue.EnqueueAsync(ToJson(pipeline), cancellationToken).ConfigureAwait(false);
 
-            // Try to save the pipeline status
-            await this.UpdatePipelineStatusAsync(pipeline, cancellationToken, ignoreExceptions: true).ConfigureAwait(false);
+            // Save the pipeline status to disk.
+            // IMPORTANT: If this fails the system will retry the "next" step stored on disk,
+            await this.UpdatePipelineStatusAsync(pipeline, cancellationToken).ConfigureAwait(false);
```
TaoChenOSU, line 206: If this fails, wouldn't it throw an uncaught exception, which would cause the whole pipeline to halt? Even if it keeps going since the messages are still there, wouldn't it continue the current step?

dotnet/CoreLib/Pipeline/DistributedPipelineOrchestrator.cs
```diff
@@ -167,28 +184,30 @@ public override async Task RunPipelineAsync(DataPipeline pipeline, CancellationT
...
+
+            // Enqueue a pointer to the pipeline (the entire pipeline doc can be too big to fit)
+            // IMPORTANT: If this fails the state on disk will be ahead, and the system will retry the step before.
+            await queue.EnqueueAsync(ToJson(new DataPipelinePointer(pipeline)), cancellationToken).ConfigureAwait(false);
```
TaoChenOSU, line 210: Why would the system retry the step before? If the previous step is successful, the message should have been removed from its queue at this point. I might be missing something here.

dotnet/CoreLib/Pipeline/DistributedPipelineOrchestrator.cs
dluc: The pipelines should never halt, thanks to the queues triggering an automatic retry in case of errors.
When we reach this line we just completed the business logic of a specific step, and we're trying to update the pipeline state stored on disk. The queue message that got us here is still in the queue, "locked". If an exception occurs, the exception bubbles up to the queue message dispatcher, where it is caught, the queue message is unlocked, and thus the same message is dispatched again, leading to a retry.

dotnet/CoreLib/Pipeline/DistributedPipelineOrchestrator.cs
dluc: When the execution gets here, the system is still processing the message dispatched. The message is still in the queue, locked. At this point the state on disk is saying "step completed" but if this code triggers an exceptions, it bubbles up to the queue dispatcher who doesn't know what went wrong, so it will just unlock the message and let it retry. The message in the queue though is about the step just completed, that's why we say the state on disk is "ahead", and there is a rollback call (line 96).

PR #73, opened 2023-10-02
Fix use of DefaultSteps from config file

dluc:

The default pipeline steps defined in appsettings.*.json and appsettings.json were not properly loaded and propagated when using DI.
Fix for https://github.com/microsoft/semantic-memory/issues/64

Fix: .NET IConfiguration doesn't support array override (when using multiple config sources): in fact arrays are merged so a shorter list in "appsettings.development.json" would not replace a longer list in "appsettings.json", but only replace the initial values. The list of default steps in appsettings.json is now empty, with defaults set via code. Values in appsettings.*.json now are properly loaded replacing the defaults.
SM config is also now stored in the service collection, so SemanticMemoryConfig can be injected via DI. Before the value passed by DI would be null, and the the pipeline orchestrator would not see settings from the config files (e.g. the list of default steps).

### DIFF
--- a/dotnet/Service/appsettings.json
+++ b/dotnet/Service/appsettings.json
@@ -34,14 +34,18 @@
       "VectorDbTypes": [
         "AzureCognitiveSearch"
       ],
+      // Note: keep the list empty in this file, to avoid unexpected merges
+      // with the list defined in appsettings.*.json.
+      // If the list is empty, SemanticMemoryConfig uses 'Constants.DefaultPipeline'.
       "DefaultSteps": [
-        "extract",
-        "partition",
-        "gen_embeddings",
-        "save_embeddings",
-        "summarize",
-        "gen_embeddings",
-        "save_embeddings"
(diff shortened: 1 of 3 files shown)


dotnet/Service/appsettings.json
```diff
@@ -34,14 +34,18 @@
       "VectorDbTypes": [
         "AzureCognitiveSearch"
       ],
+      // Note: keep the list empty in this file, to avoid unexpected merges
```
TaoChenOSU, line 37: nit: Could we mention this in a trouble shooting guide somewhere with examples? We have arrays in other places in the settings.

TaoChenOSU (approved):

dotnet/Service/appsettings.json
dluc: I'll try to think about a place where we could add similar notes. It's something about .NET and not specific to this project, e.g. some people already know about this - I left the note just in case one wonders why the default is empty

PR #76, opened 2023-10-04
Allow to share memory dependencies with external applications e.g. ASP.NET

dluc:

When an application like ASP.NET shares with the memory builder the local service collection, the user might want all the memory dependencies to be injected into it. 

E.g. methods like `WithCustomImageOcr` inject dependencies into the service collection used by the memory, but not the one used by the end user application:

```csharp

var builder = WebApplication.CreateBuilder(args);

ISemanticMemoryClient memory = new MemoryClientBuilder(builder.Services)
    .WithOpenAIDefaults(Env.Var("OPENAI_API_KEY"))
    .WithCustomImageOcr(new MyOcrEngine())
    .Build();
```

and a controller won't see the OCR engine, unless that is registered also in builder.Services

```csharp
[ApiController]
[Route("[controller]")]
public class MemoryController : ControllerBase
{
  private readonly IOcrEngine _ocr;

  public SomeController(IOcrEngine ocr)
  {
      this._ocr = ocr;
  }
}
```

Add a new class to proxy `WithXYZ` calls to multiple service collection instances, e.g. the one used by the MemoryBuilder and the one used by ASP.NET and passed to the builder ctor.

### DIFF
--- a/dotnet/CoreLib/AppBuilders/MemoryClientBuilder.cs
+++ b/dotnet/CoreLib/AppBuilders/MemoryClientBuilder.cs
@@ -9,6 +9,7 @@ using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.SemanticKernel.AI.Embeddings;
 using Microsoft.SemanticMemory.AI;
+using Microsoft.SemanticMemory.AppBuilders;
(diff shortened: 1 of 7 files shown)


dluc:
Should fix https://github.com/microsoft/semantic-memory/issues/67

TaoChenOSU (approved):

PR #78, opened 2023-10-04
NL2SQL Update default embedding model

crickman:

Update default emedding model to `ada_002`

`ada_003` not yet publicly available


dluc (approved):

dluc:
Closes https://github.com/microsoft/semantic-memory/issues/77

PR #81, opened 2023-10-06
Transition to Polly.Core

anthonypuppo:
The `Polly.Core` package was recently released and is recommended for new projects (see [getting started](https://www.pollydocs.org/getting-started.html) and [migration guide](https://www.pollydocs.org/migration-v8.html)).

Update to `Polly.Core` and transition from using `Policy` (v7) to `ResiliencePipeline` (v8).


dotnet/CoreLib/DataFormats/WebPages/WebScraper.cs
```diff
@@ -47,7 +46,11 @@ private async Task<Result> GetAsync(Uri url)
 
         // TODO: perf/TCP ports/reuse client
         using var client = new HttpClient();
-        HttpResponseMessage? response = await RetryLogic().ExecuteAsync(() => client.GetAsync(url)).ConfigureAwait(false);
+        HttpResponseMessage? response = await RetryLogic().ExecuteAsync(async (cancellationToken) =>
+        {
+            var response = await client.GetAsync(url, cancellationToken: cancellationToken).ConfigureAwait(false);
+            return response;
+        }).ConfigureAwait(false);
```
dluc, line 53: the code seems much more verbose and complex than before, any reason? any way to simplify?

anthonypuppo: It's because `ExecuteAsync` was changed to pass a callback that returns `ValueTask<TResult>`. Compiler was unhappy trying to convert from `Task<TResult>`. Thought it looked more readable expanded to a block, but the below would accomplish the same thing on one line (let me know and I'll make the change):

```csharp
HttpResponseMessage? response = await RetryLogic().ExecuteAsync(async (cancellationToken) => await client.GetAsync(url, cancellationToken).ConfigureAwait(false)).ConfigureAwait(false);
```

dluc: yes that looks better, removing extra brackets and indenting should look like:

```csharp
HttpResponseMessage? response = await RetryLogic()
            .ExecuteAsync(async cancellationToken => await client.GetAsync(url, cancellationToken).ConfigureAwait(false))
            .ConfigureAwait(false);
```

anthonypuppo: Updated.

dluc (approved):

PR #86, opened 2023-10-11
Update Azure.Search.Documents to 11.5.0-beta.5

anthonypuppo:
This release contains multiple changes that Semantic Memory can take advantage of. Also aligns with the version used by Semantic Kernel which will prevent runtime errors during the next SK release due to breaking changes.

See [changelog](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/search/Azure.Search.Documents/CHANGELOG.md#1150-beta5-2023-10-09).

Update Azure.Search.Documents to 11.5.0-beta.5 and fix associated regressions.


dotnet/CoreLib/MemoryStorage/AzureCognitiveSearch/AzureCognitiveSearchMemory.cs
```diff
@@ -109,16 +109,16 @@ public async Task<string> UpsertAsync(string indexName, MemoryRecord record, Can
 
         var client = this.GetSearchClient(indexName);
 
-        SearchQueryVector vectorQuery = new()
+        RawVectorQuery vectorQuery = new()
         {
             KNearestNeighborsCount = limit,
-            Value = embedding.Data.ToArray(),
+            Vector = embedding.Data.ToArray(),
             Fields = { AzureCognitiveSearchMemoryRecord.VectorField }
```
dluc, line 116: could you add `Exhaustive = true`?

we might make it configurable in future, but for now I'd like having the best matches by default

dotnet/CoreLib/MemoryStorage/AzureCognitiveSearch/AzureCognitiveSearchMemory.cs
```diff
@@ -458,7 +458,7 @@ private SearchIndex PrepareIndexSchema(string indexName, VectorDbSchema schema)
                         IsFacetable = false,
                         IsSortable = false,
                         VectorSearchDimensions = field.VectorSize,
-                        VectorSearchConfiguration = VectorSearchConfigName,
+                        VectorSearchProfile = VectorSearchConfigName,
```
dluc, line 461: nit: rename `VectorSearchConfigName` to `VectorSearchProfileName`

anthonypuppo:
> could you add `Exhaustive = true`?
> 
> we might make it configurable in future, but for now I'd like having the best matches by default

> nit: rename `VectorSearchConfigName` to `VectorSearchProfileName`

Done.

dluc (approved):

dluc (changes requested):
There's some problem, the functional tests are failing:

```
he request is invalid. Details: definition : The field 'embedding' uses a vector search algorithm configuration '' which is not defined.
Status: 400 (Bad Request)
ErrorCode: InvalidRequestParameter

Content:
{"error":{"code":"InvalidRequestParameter","message":"The request is invalid. Details: definition : The field 'embedding' uses a vector search algorithm configuration '' which is not defined.","details":[{"code":"UnknownVectorAlgorithmConfiguration","message":"The field 'embedding' uses a vector search algorithm configuration '' which is not defined. Parameters: definition"}]}}
```

anthonypuppo:
Should be good now. In the latest `2023-10-01-Preview` API version `vectorSearchProfile` replaced `vectorSearchConfiguration`. The index now gets created with the appropriate profile that the vector fields can reference.

https://github.com/microsoft/semantic-memory/pull/86/files#diff-76900a949ad74d153b00b8ebdbe4bebc905d3b952a270b171e8e383c28c273cbL429

dotnet/CoreLib/MemoryStorage/AzureCognitiveSearch/AzureCognitiveSearchMemory.cs
```diff
@@ -419,16 +420,21 @@ private SearchIndex PrepareIndexSchema(string indexName, VectorDbSchema schema)
 
         indexName = this.NormalizeIndexName(indexName);
 
-        const string VectorSearchConfigName = "SemanticMemoryDefaultCosine";
+        const string VectorSearchProfileName = "SemanticMemoryDefaultHnsw";
+        const string VectorSearchHnswConfig = "SemanticMemoryDefaultHnswConfig";
```
anthonypuppo, line 424: FYI changed these values to better reflect the new profile/algorithm definitions. Pointing it out in case you have a better naming scheme.

anthonypuppo:
FWIW I can also do #92 as part of this PR, initially scoped it down to just resolving breaking changes.

dotnet/CoreLib/MemoryStorage/AzureCognitiveSearch/AzureCognitiveSearchMemory.cs
dluc: have you tested if changing these strings is a breaking change? I think we need to add some documentation answering this questions:

* what if I created an index using different profile name?
* what if I created an index using different name for the search algo?
* why does the search algo have a name? can I use Cosine similarity twice with two different name?
* can I have multiple profiles? what would be the use case?

I suspect there's going to be a lot of "configuration" data to be passed in as soon as we add support for custom indexes.

Since we're going to change these strings now, probably breaking things, could you go with these values for now:

```
const string VectorSearchProfileName = "Default";
const string VectorSearchHnswConfigName = "Default";
```

dluc: I'll need to do some tests before merging, understanding better these changes and the implications of `Exhaustive`, going through https://learn.microsoft.com/en-us/azure/search/vector-search-how-to-create-index?tabs=config-2023-10-01-Preview%2Crest-2023-10-01-Preview%2Cpush%2Cportal-check-index and trying a combination of old code migrating to new code, old indexes being used by the new code, etc

anthonypuppo: From my (albeit limited) testing of the new API version these are the relevant changes from a SM perspective:

- Vector field profiles. These are defined as part of the index creation. The only thing that can really be configured right now is the nearest-neighbor algorithm used (so cognitive search can index appropriately) but it's probably setting up for future settings. Each vector field needs to reference a profile (previously, they just referenced the algorithm).
- New algorithm, `ExhaustiveKNN`, which (as the name implies) always uses an exhaustive vector search for queries.
- The `Exhaustive` search query flag. If HNSW is used for a field, you can switch to exhaustive KNN during a query, but not vice versa, as the necessary data structures for HNSW won't be available.

To answer your questions (again, my from limited understanding):

- `what if I created an index using different profile name?` No impact, profiles weren't available in prior API versions.
- `what if I created an index using different name for the search algo?` No impact, this is just so cognitive search can setup the index appropriately.
- `why does the search algo have a name? can I use Cosine similarity twice with two different name?` This was so the vector fields could reference an algorithm, same concept as above.
- `can I have multiple profiles? what would be the use case?` Yup, purpose will probably become clearer as the API is built out more, but for right now it's just for using different algorithms.

In short, pre-existing indexes shouldn't be impacted as their structure was already defined (HNSW). The existing default profile/algorithm (HNSW) takes care of the 99% use case imo. Defaulting `Exhaustive` is probably a good idea for now until the ACS methods support a way for the caller to configure the search options a bit more.

dluc:
There seems to be some problem at the query level.

How to repro:
* Setup Azure Cognitive Search
* Delete all indexes
* Run example 001-dotnet-Serverless

Code in main branch:
> Question: What's Semantic Kernel?
> Answer: Semantic Kernel is a lightweight SDK [...]

Code in the PR:
> Question: What's Semantic Kernel?
> Answer: INFO NOT FOUND

anthonypuppo:
> There seems to be some problem at the query level.
> 
> How to repro:
> 
> * Setup Azure Cognitive Search
> * Delete all indexes
> * Run example 001-dotnet-Serverless
> 
> Code in main branch:
> 
> > Question: What's Semantic Kernel?
> > Answer: Semantic Kernel is a lightweight SDK [...]
> 
> Code in the PR:
> 
> > Question: What's Semantic Kernel?
> > Answer: INFO NOT FOUND

Hmm, I'm not able to reproduce. Tested on a branch new ACS resource just to be safe. The output I'm getting more or less matches what's expected as described by the comment at the end of the main program file.

The only thing I'm changing about the example is uncommenting the `WithAzureCognitiveSearch` call. It looks like it's uploading things fine as removing the purge logic I can query the data fine in the Azure portal. If you do the same and remove the purge logic, can you also query it in the portal?

anthonypuppo:
@dluc To follow up on this, I've also been seeing intermittent issues where the logic expects the indexed content to be immediately available but isn't until a few seconds after. 

From the [MS docs](https://learn.microsoft.com/en-us/azure/search/search-get-started-text?tabs=dotnet) (load a document section):
> ```The 2-second delay compensates for indexing, which is asynchronous, so that all documents can be indexed before the queries are executed. Coding in a delay is typically only necessary in demos, tests, and sample applications.```

It may be worth adding a test specific retry policy to attempt retrieving the expected result a few times. It doesn't have to be complicated (ie. 1 -> 3 -> 10 seconds) but should alleviate the occasional false negative.

dluc (approved):
I've done another round of tests, creating data with the code in main branch and querying it with the code in this PR, and everything works fine. Found some outdated code that I fixed, and added some extra debugging facility in a couple of files, in case of future failures to investigate.

PR #87, opened 2023-10-11
New API to delete indexes

dluc:

New API allowing apps to delete entire indexes.

New API in `ISemanticMemoryClient` to delete indexes.
New HTTP method allowing to delete indexes.
New handler responsible for deleting vectors and files in storage.
New logic allowing a pipeline not having a document ID (just the index name to delete).
New logic to delete storage directories after a pipeline is complete and the status file has been written.
New functional tests (with manual verifications).
Fix OpenAPI signature of DELETE methods.

### DIFF
--- a/dotnet/CoreLib/ContentStorage/AzureBlobs/AzureBlobsStorage.cs
+++ b/dotnet/CoreLib/ContentStorage/AzureBlobs/AzureBlobsStorage.cs
@@ -107,6 +107,10 @@ public class AzureBlobsStorage : IContentStorage
         string index,
         CancellationToken cancellationToken = default)
     {
+        // Note: Azure Blobs doesn't have an artifact for "directories", which are just a detail in a blob name
+        //       so there's no such thing as creating a directory. When creating a blob, the name must contain
+        //       the directory name, e.g. blob.Name = "dir1/subdir2/file.txt"
+
         this._log.LogTrace("Creating container '{0}' ...", this._containerName);
 
         await this._containerClient
@@ -122,6 +126,10 @@ public class AzureBlobsStorage : IContentStorage
         string documentId,
         CancellationToken cancellationToken = default)
     {
+        // Note: Azure Blobs doesn't have an artifact for "directories", which are just a detail in a blob name
+        //       so there's no such thing as creating a directory. When creating a blob, the name must contain
+        //       the directory name, e.g. blob.Name = "dir1/subdir2/file.txt"
+
         this._log.LogTrace("Creating container '{0}' ...", this._containerName);
 
         await this._containerClient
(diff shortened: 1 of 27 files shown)


dotnet/CoreLib/ContentStorage/AzureBlobs/AzureBlobsStorage.cs
```diff
@@ -285,6 +291,44 @@ private async Task<long> InternalWriteAsync(string directoryName, string fileNam
...
+            {
+                // Skip root and empty names
+                if (string.IsNullOrWhiteSpace(blob.Name) || blob.Name == prefix) { continue; }
+
+                // Remove the prefix, skip root and empty names
+                var fileName = blob.Name.Trim('/').Substring(prefix.Trim('/').Length).Trim('/');
+                if (string.IsNullOrWhiteSpace(fileName)) { continue; }
+
+                // Don't delete the pipeline status file
+                if (fileName == Constants.PipelineStatusFilename) { continue; }
```
TaoChenOSU, line 316: Why not delete the pipeline status file?

TaoChenOSU:
What happens if a pipeline is still running and there is a delete request that tries to delete an index?

dotnet/CoreLib/ContentStorage/AzureBlobs/AzureBlobsStorage.cs
dluc: deleting the file while the pipeline is in execution would break the pipeline in case of a retry. The file is updated "after" the deletion step is executed. And the corresponding message from the queue is removed "after" the step is executed. If the message remains in the queue due to some error, it would retry, not find the file, and start logging errors about a corrupted document.

dluc:
> What happens if a pipeline is still running and there is a delete request that tries to delete an index?

A lot of things could happen, depending on the order. It's something we should look into, and a quick thought would be introducing an "index lock" for the deletion step, so that the deletion won't start until it has acquired the lock. Similar for documents deletions. There's a lot of work to do in this area, that will surface once we start working with chats, e.g. a document being repeatedly updated.
There are also other considerations, e.g. creating and deleting indexes in some storages like Pinecone can take minutes, so there's a lot of concurrency problems that could arise.
For now I would say not to abuse of this API, don't delete indexes too often, until we mature the API and the approach to concurrent writes.

dotnet/CoreLib/ContentStorage/AzureBlobs/AzureBlobsStorage.cs
TaoChenOSU: nit: should we document this somewhere?

TaoChenOSU:
> > What happens if a pipeline is still running and there is a delete request that tries to delete an index?
> 
> A lot of things could happen, depending on the order. It's something we should look into, and a quick thought would be introducing an "index lock" for the deletion step, so that the deletion won't start until it has acquired the lock. Similar for documents deletions. There's a lot of work to do in this area, that will surface once we start working with chats, e.g. a document being repeatedly updated. There are also other considerations, e.g. creating and deleting indexes in some storages like Pinecone can take minutes, so there's a lot of concurrency problems that could arise. For now I would say not to abuse of this API, don't delete indexes too often, until we mature the API and the approach to concurrent writes.

Say if we have a pipeline X that contains 5 steps. A delete request comes in right after step 3, which will create another pipeline, say Y. What will be the execution order? Seems like it'd be an arbitrary order even with an index lock. If we wait until the pipeline X finishes before Y, that seems like a waste of resource. Is there a way to stop pipeline X?

dluc:
> Say if we have a pipeline X that contains 5 steps. A delete request comes in right after step 3, which will create another pipeline, say Y. What will be the execution order? Seems like it'd be an arbitrary order even with an index lock. If we wait until the pipeline X finishes before Y, that seems like a waste of resource. Is there a way to stop pipeline X?

We don't control the underlying queueing platform, so for instance we don't have a guarantee about the order X and Y will run. Assuming the worst, pipeline X will enter an error state because the index disappeared, it might recreate the vector index without the storage index or viceversa. In any case, the index will be broken, unless we design a locking system, thinking about all the scenarios. I don't think it's about wasted resources, but about having unclean state.  We can think about blocking other pipelines, anyway, we're talking about complex concurrent write operations across multiple storage solutions (vectors and files) so it's not something we can easily fix now. I would go ahead accepting the limitation, and design for this problem separately.

There are multiple situations where we will encounter concurrent writes:
* index deletions
* document deletions
* document updates
* memory consolidation
so there's a lot to consider. We're still on a preview version so I'd say let's move forward by incremental improvements.

dotnet/CoreLib/ContentStorage/AzureBlobs/AzureBlobsStorage.cs
dluc: yes, see `CleanUpAfterCompletionAsync`

PR #103, opened 2023-10-19
Add basic PowerPoint text extraction

anthonypuppo:
Adds support for extracting text content from PowerPoint (.pptx) documents.

Uses the Open XML SDK for text extraction.


crickman:
This is great, thank you.  Its been on the back of my mind for a while that PPT support was a gap that needed filling here.

dluc (approved):

PR #104, opened 2023-10-19
Add basic Excel text extraction

anthonypuppo:
Adds support for extracting text content from Excel (.xlsx) documents.

Uses [ClosedXML](https://github.com/ClosedXML/ClosedXML) for text extraction.


dotnet/CoreLib/DataFormats/Office/MsExcelDecoder.cs
```diff
@@ -0,0 +1,59 @@
...
+                for (var i = 0; i < cellsUsed.Count; i++)
+                {
+                    var cell = cellsUsed[i];
+                    var cellValue = cell.Value.ToString(CultureInfo.CurrentCulture);
+
+                    sb.Append(cell.Value);
+
+                    if (i < cellsUsed.Count - 1)
+                    {
+                        sb.Append(' ');
```
anthonypuppo, line 49: Ideally this would be configurable depending on if the caller knows the type of data being extracted (ie. enclose individual cell values in quotations and separate with a comma) but for now this single space separation should be sufficient.

dluc:
Could you address the conflicts please? 
Could you update the PR description explaining how complex spreadsheets are handled and known limitations? E.g. think about multiple sheets, external content, scripting, etc.

dotnet/CoreLib/DataFormats/Office/MsExcelDecoder.cs
```diff
@@ -0,0 +1,59 @@
...
+            foreach (var row in rowsUsed)
+            {
+                var cellsUsed = row.CellsUsed().ToList();
+
+                for (var i = 0; i < cellsUsed.Count; i++)
+                {
+                    var cell = cellsUsed[i];
+                    var cellValue = cell.Value.ToString(CultureInfo.CurrentCulture);
+
+                    sb.Append(cell.Value);
```
dluc, line 45: should this append `cellValue` defined at line 43 instead?

dotnet/CoreLib/DataFormats/Office/MsExcelDecoder.cs
dluc: fixed

dotnet/CoreLib/DataFormats/Office/MsExcelDecoder.cs
dluc: fixed

dluc (approved):

PR #105, opened 2023-10-19
Add support for listing memory contents (issue 69)

ogkent:

Related to issue https://github.com/microsoft/semantic-memory/issues/69.

**Goal:** we want to be able to inspect the contents of the memory store

Opening as a draft PR as I haven't discussed this change with the maintainers. I expect there are many problems with it, but thought it would be good to open to start a conversation.

I'm exposing the internal `ListAsync` function that already exists in the db.


ogkent:
@microsoft-github-policy-service agree company="Microsoft"

ogkent:
I don't think we need this right now. We can find alternative ways to get the contents of memory.

PR #106, opened 2023-10-19
Renaming Semantic Memory to Kernel Memory (non breaking changes)

dluc:

Renaming the solution to the official name: Kernel Memory

This is the first PR, containing only non breaking changes, such as docs, links and such.
One or more PR will follow with the breaking changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
 # Kernel Memory
 
-**Kernel Memory** (SM) is an open-source [service](dotnet/Service/README.md) and
+**Kernel Memory** (KM) is an open-source [service](dotnet/Service/README.md) and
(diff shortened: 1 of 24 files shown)


alexchaomander (dismissed):

README.md
```diff
@@ -1,22 +1,22 @@
-# Semantic Memory
+# Kernel Memory
 
-**Semantic Memory** (SM) is an open-source [service](dotnet/Service/README.md) and
+**Kernel Memory** (SM) is an open-source [service](dotnet/Service/README.md) and
```
TaoChenOSU, line 3: Do we still want to use "SM" as the abbreviation?

dluc: KM, thanks!

TaoChenOSU (approved):

PR #111, opened 2023-10-20
Add basic Outlook text extraction

anthonypuppo:
Adds support for extracting text content from Outlook (.msg) documents.

Uses [MSGReader](https://github.com/Sicos1977/MSGReader) for text extraction.


crickman:
Thank you for this.

I do wonder if live connection via a plug-in/function might be a more common approach for accessing email.  

I suppose if an important .msg needed to be imported...why not?

anthonypuppo:
> Thank you for this.
> 
> 
> 
> I do wonder if live connection via a plug-in/function might be a more common approach for accessing email.  
> 
> 
> 
> I suppose if an important .msg needed to be imported...why not?

@crickman That would probably be way more ideal from nearly every angle (UX, realtime data, etc). However, a connection to graph might not be high on a team's priority list, so this adds support for those "good for now" cases 😄.

dluc:
Could you address the conflicts please?

Longer term I think we'll add MSGraph, however in many companies MSGraph API is blocked, IMAP too, so it's nice having some solution :-)

anthonypuppo:
Closing due to how stale this branch has gotten.

PR #114, opened 2023-10-21
Fix/refactor minRelevance

### LINKED ISSUE #109: Setting a Filter with MemoryFilter.MinRelevance value, makes the Az Search request fail.
I don't know if this is something we can do in Az search, but I'd like to get documents with Score/Relevance greater than a given value.

The ```SearchAsync``` (and AskAsync) has a Filter parameter, and there there is a ```MinRelevance``` property. However, this property is not used later in code, and that generates an Azure Search API request with the Filter parameter empty, so it fails. This snippet fails:

```csharp
        var data = await _memory.SearchAsync(
            query: _question,
            index: Constants.AzureSearchIndexName,
            filter: new MemoryFilter { MinRelevance = 0.8f }
            limit: 10
        );
```

I'd like to have that Filter by Score, cos I'm having the following behaviour:
I have indexed just 3 documents, 2 of them are related to Domain Driven Design, and the 3rd one is the wikipedia into about Spiderman. I ask then the following question to the ```AskAsync``` method: _"When an Entity should be considered an aggregate in DDD"_ 

In the relevant results, I'm getting Citations/Parts from the Spiderman document. Obviously, they are sorted below the ones related with DDD, but as when the Prompt is composed, facts are added depending on the Tokens size, you can end up sending facts from Spiderman document. If we can filter the Search query to say "only search documents having score > 0.8" will avoid sending Spiderman facts to the Prompt.

anthonypuppo:
Refactor `minRelevance` logic:

- Remove `MemoryFilter.MinRelevance`
- Add optional argument `minRelevance` to `SearchAsync` and `AskAsync` and forward as necessary

Fixes #109.

Remove the unused `MemoryFilter.MinRelevance` property and instead follow the optional argument pattern.

BREAKING change.

### DIFF
--- a/dotnet/ClientLib/IKernelMemory.cs
+++ b/dotnet/ClientLib/IKernelMemory.cs
@@ -168,8 +168,9 @@ public interface IKernelMemory
     /// <param name="index">Optional index name</param>
     /// <param name="filter">Filter to match</param>
     /// <param name="filters">Filters to match (using inclusive OR logic). If 'filter' is provided too, the value is merged into this list.</param>
-    /// <param name="cancellationToken">Async task cancellation token</param>
     /// <param name="limit">Max number of results to return</param>
+    /// <param name="minRelevanceScore">Minimum Cosine Similarity required</param>
+    /// <param name="cancellationToken">Async task cancellation token</param>
     /// <returns>Answer to the query, if possible</returns>
     public Task<SearchResult> SearchAsync(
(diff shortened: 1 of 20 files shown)


anthonypuppo:
Marking as a draft just until there's some confirmation using the optional argument pattern for this is the way to go.

dluc (dismissed):

PR #115, opened 2023-10-21
ACS only add filter if the conditions are non-empty

anthonypuppo:
If a `MemoryFilter` is passed to the ACS client without any filters it creates an invalid query string `()`.

Helps address #109.

Only append the condition filter is filters actually exist.


dotnet/CoreLib/MemoryStorage/AzureCognitiveSearch/AzureCognitiveSearchMemory.cs
```diff
@@ -576,9 +576,13 @@ private static string BuildSearchFilter(ICollection<MemoryFilter> filters)
                 {
                     var fieldValue = keyValue.Value?.Replace("'", "''", StringComparison.Ordinal);
                     return $"tags/any(s: s eq '{keyValue.Key}{Constants.ReservedEqualsSymbol}{fieldValue}')";
-                });
+                })
+                .ToList();
 
-            conditions.Add($"({string.Join(" and ", filterConditions)})");
+            if (filterConditions is { Count: > 0 })
```
dluc, line 582: how does the code produce an empty `filterConditions` list from a non empty `filters`?

dluc: all calls to BuildSearchFilter are gated by a `if (filters is { Count: > 0 })` (line) so this function is always passed at least one filter. I can't see a scenario where `filterConditions` would be empty.

could you include a test to reproduce the error?

anthonypuppo: This can happen if an empty instance of `MemoryFilter` is passed (see my comment [here](https://github.com/microsoft/kernel-memory/issues/109#issuecomment-1772800062)). The `string.Join` call assumes `filterConditions` is a non-empty collection which, when it is, just produces `()`. The current scenario is when setting `MemoryFilter.MinRelevance` without any filters being configured (see #109).

The scenario shouldn't really exist _but_ is still possible, so this change protects against it.

dluc: is it possible to add a test to reproduce the error?

tests/FunctionalTests/ServerLess/FilteringTest.cs
```diff
@@ -52,8 +52,13 @@ public async Task ItSupportsASingleFilter(string memoryType)
             await Task.Delay(TimeSpan.FromSeconds(2));
         }
 
+        // Simple filter: empty filter returns the memory
+        var answer = await this._memory.AskAsync("What is Orion?", filter: new(), index: indexName);
```
anthonypuppo, line 56: @dluc See here for a functional test. I assume the expected response if an empty filter is passed is to assume no filter should apply to the query?

Before the fix you'd get the following error when using ACS:
```
Azure.RequestFailedException : Invalid expression: Expression expected at position 1 in '()'.
Parameter name: $filter
Status: 400 (Bad Request)
```

dluc (approved):

PR #135, opened 2023-11-04
Refactor SHA256 in CoreLib

anthonypuppo:
Optimize the hashing logic in some of the pipeline handlers.

- Add `BinaryDataExtensions.ToSHA256`
- Refactor the SHA256 calculation to use `ToMemory` on the `BinaryData` object which removes an array allocation


crickman:
Nice touch!  Thank you.

dluc (approved):

Issue #131, opened 2023-11-01
Slow document processing


KSemenenko:
Hello is there a way to pralellize RunPipelineAsync methid to execute steps is paralel? 
like summarization or upload differnt part of documets?

KSemenenko:
all time take step second "gen_embeddings" after summarization

KSemenenko:
<img width="364" alt="Screenshot 2023-11-06 at 22 04 57" src="https://github.com/microsoft/kernel-memory/assets/4385716/471b1938-45f4-4fd3-9976-e18a129beabf">

dluc:
hi @KSemenenko pipeline steps cannot run in parallel because each step depends on the previous one, however, if you import multiple documents, each document is processed in parallel. If you use the async pipelines, eg with Azure Queue or RabbitMQ, the ingestion runs in the background, without blocking your apps. 

Summarizing still takes time, but it runs after indexing the document chunks, so you shouldn't see much difference.

In the latest nuget the summarization step is disabled by default, so you might see an improvement, unless you still want to use summaries.

KSemenenko:
@dluc  I've made a PR and I think these files can definitely be run in parallel.
what do you think?

dluc:
We have removed Summarization from the default pipeline, so by default ingestion is quite fast now. About the PR, I think the best approach is using that as a custom handler, rather than changing the existing one. We will provide soon some examples showing how to plug in custom handlers, in the service and in the serverless option.

KSemenenko:
I did as you suggest. 

Anyway embedding is slow.  I’m increase number of tokens but it’s still not that fast as some services.

that PR is more like code example for discussion.

KSemenenko:
@dluc 
[National Planning Policy Framework.pdf](https://github.com/microsoft/kernel-memory/files/13420014/National.Planning.Policy.Framework.pdf)
you can also find it here https://www.gov.uk/government/publications/national-planning-policy-framework--2
I have my favorite test pdf, and for me processign takes like forever

KSemenenko:
also it will be good to see progress for processing, somehome

dluc:
@KSemenenko about seeing progress, the status endpoint returns `DataPipelineStatus` which allows to see the pipeline progress. See `completed`, `steps`, `remaining_steps` and `completed_steps`. The same feature is used by `IKernelMemory.IsDocumentReadyAsync()`

For instance:

GET http://localhost:9001/upload-status?documentId=doc001

```json
{
[...]

PR #147, opened 2023-11-09
4 thread in parallel

KSemenenko:
Relate to #131, Im trying to seepd up document processing

just paralell tasks for now

so PR is about do discuss this idea to improve perfomance


dotnet/CoreLib/Handlers/GenerateEmbeddingsHandler.cs
```diff
@@ -60,22 +60,25 @@ public class GenerateEmbeddingsHandler : IPipelineStepHandler
         {
             // Track new files being generated (cannot edit originalFile.GeneratedFiles while looping it)
             Dictionary<string, DataPipeline.GeneratedFileDetails> newFiles = new();
+            var throttler = new SemaphoreSlim(initialCount: 4);
```
anthonypuppo, line 63: nit: A more sensible default may be `Environment.ProcessorCount` so it scales according to the compute offered by the machine.

dotnet/CoreLib/Handlers/GenerateEmbeddingsHandler.cs
```diff
@@ -138,18 +141,23 @@ public class GenerateEmbeddingsHandler : IPipelineStepHandler
                                 ArtifactType = DataPipeline.ArtifactTypes.TextEmbeddingVector
                             };
                             embeddingFileNameDetails.MarkProcessedBy(this);
-                            newFiles.Add(embeddingFileName, embeddingFileNameDetails);
+                            lock (newFiles)
```
anthonypuppo, line 144: The `lock` seems a bit unnecessary here. Making `newFiles` a `ConcurrentDictionary` would accomplish the thread safety.

FYI don't lock on shared instances as it opens up the potential for contention/deadlocks. Have a separate object instance that you can lock on only for that purpose.

dotnet/CoreLib/Handlers/SummarizationHandler.cs
```diff
@@ -95,27 +97,33 @@ public class SummarizationHandler : IPipelineStepHandler
                             var destFile = uploadedFile.GetHandlerOutputFileName(this);
                             await this._orchestrator.WriteFileAsync(pipeline, destFile, new BinaryData(summary), cancellationToken).ConfigureAwait(false);
 
-                            summaryFiles.Add(destFile, new DataPipeline.GeneratedFileDetails
+                            lock (summaryFiles)
```
anthonypuppo, line 100: See my comment [here](https://github.com/microsoft/kernel-memory/pull/147/files#r1388085382).

dotnet/CoreLib/Handlers/GenerateEmbeddingsHandler.cs
KSemenenko: It can be not expected, if for example you have a server with 64 processors, and you have 100 files parallel jobs for process files, then maybe 6400 requests in ada will start to fail on TooManyRequests. I think to make it configurable, what do you think?

dotnet/CoreLib/Handlers/GenerateEmbeddingsHandler.cs
KSemenenko: It's a good idea, but ConcuretDictirany is much slower than just dictionary and lock
https://github.com/dotnet/runtime/issues/667

let me fix lock issue

dotnet/CoreLib/Handlers/SummarizationHandler.cs
KSemenenko: fix, but I did this lock, becase this dictionary is private end exisit only in this method, but you are right. it's safe to have separte variable for this

dotnet/CoreLib/Handlers/GenerateEmbeddingsHandler.cs
KSemenenko: Is it good idea add int DataPipeline propery for number of parallel tasks? so we can control it

dluc: Isn't the work IO bound, in which case the number of processors doesn't really matter?

dluc:
The pipeline scalability is based on asynchronous queues being processed in parallel. If a message in the queue is taking too long because it is doing too much work and the work could be divided, why not leverage the existing infrastructure and split the task over multiple messages?

anthonypuppo:
> About embedding, many embedding generators support passing a list of strings to generate multiple embeddings at once. Maybe we should look into that too?

@dluc FWIW I have an open PR in the SK repo to enable batching microsoft/semantic-kernel#3295. If SK supports batching natively the changes required here are minimal. Or, could create a KM specific `ITextEmbedding` interface and change the default implementation (would mean rewriting pre-existing implementations that SK has though like OpenAI, HuggingFace, etc).

dotnet/CoreLib/Handlers/GenerateEmbeddingsHandler.cs
KSemenenko: if we will have a lot of tasks they will case Thread starvation, 
anyway most importat there is 
``` cs
 (string summary, bool success) = await this.SummarizeAsync(content).ConfigureAwait(false);
```

KSemenenko: we have to check it wih many files wich prosessed at the same time and I added Environment.ProcessorCount

anthonypuppo: > It can be not expected, if for example you have a server with 64 processors, and you have 100 files parallel jobs for process files, then maybe 6400 requests in ada and it will start to fail on TooManyRequests. I think to make it configurable, what do you think?

IMO configurable is ideal, but I'd argue the resilience should be handled by the underlying provider and not the handler itself.

KSemenenko:
> The pipeline scalability is based on asynchronous queues being processed in parallel. If a message in the queue is taking too long because it is doing too much work and the work could be divided, why not leverage the existing infrastructure and split the task over multiple messages?
> 
> About embedding, many embedding generators support passing a list of strings to generate multiple embeddings at once. Maybe we should look into that too?

As I see in my tests, this `uploadedFile.GeneratedFiles` is small files are part of big one, so I need somehow imporove perfomance.

dluc:
Here's my suggestion: rather than changing `GenerateEmbeddingsHandler`, create a new handler, e.g. `GenerateEmbeddingsInParallelHandler`, assign a name e.g. `gen_embeddings_parallel`, and select this handler in your deployments. You can select handlers while inserting data (see `steps` param) or configure the service to use your custom handlers by default.

Longer term, what I would recommend investigating: 
* currently, after extracting text from a file, and after partitioning the text in multiple chunks, only 1 message is queued, moving from "partition" step to the next step "gen_embeddings"
* rather than enqueuing only 1 message, enqueue N messages, one for every partition. This solution would allow to process all partitions in parallel over N machines, leveraging the existing pub-sub infra.

Pros and Cons:
* Pros: infinite scaling, ignoring individual nodes specs (e.g. number of cores).
* Cons: code changes to the core of the pipeline, vs changes to a single handler. E.g. ensuring N messages are enqueued at least once, aka increased logic to be resilient.

KSemenenko:
@dluc  btw how can I add my custim handler? I don't see any extestions like `WithCustomHandler`

dluc:
> @dluc btw how can I add my custim handler? I don't see any extestions like `WithCustomHandler`

I was about to provide an example, but the code is too complex. Currently there's two different methods, one `AddHandler` in the `Memory` class (if you are using the serverless memory) and a `AddHandlerAsync` in the orchestrator (if you are using the service). A bit too hard to setup, needs some work to follow the usual builder approach.

KSemenenko:
@dluc  what do you think about ```WithCustomHandler``` I added in this PR?

KSemenenko:
@dluc  I tested latest version and it works fine with perfomncae, so maybe we can convert this PR into "custom handler extenstions" ?

extensions/LlamaSharp/LlamaSharp.FunctionalTests/LlamaSharp.FunctionalTests.csproj
```diff
@@ -29,7 +29,7 @@
     </ItemGroup>
 
     <ItemGroup>
-      <ProjectReference Include="..\..\..\service\tests\TestHelpers\TestHelpers.csproj" />
+        <ProjectReference Include="..\..\..\service\tests\TestHelpers\TestHelpers.csproj"/>
```
dluc, line 32:
```suggestion
     <ProjectReference Include="..\..\..\service\tests\TestHelpers\TestHelpers.csproj" />
```

service/Core/Pipeline/InProcessPipelineOrchestrator.cs
```diff
@@ -166,8 +167,11 @@ public void AddHandler(IPipelineStepHandler handler)
     ///<inheritdoc />
     public override async Task RunPipelineAsync(DataPipeline pipeline, CancellationToken cancellationToken = default)
     {
+        var sw = Stopwatch.StartNew();
```
dluc, line 170:
```suggestion
```

service/Core/Pipeline/InProcessPipelineOrchestrator.cs
```diff
@@ -166,8 +167,11 @@ public void AddHandler(IPipelineStepHandler handler)
...
         // Files must be uploaded before starting any other task
         await this.UploadFilesAsync(pipeline, cancellationToken).ConfigureAwait(false);
+        sw.Stop();
+        this.Log.LogInformation("Pipeline '{0}/{1}' UploadFiles Time:{2}", pipeline.Index, pipeline.DocumentId, sw.Elapsed);
```
dluc, line 174:
```suggestion
```

service/Core/Pipeline/InProcessPipelineOrchestrator.cs
```diff
@@ -180,10 +184,14 @@ public override async Task RunPipelineAsync(DataPipeline pipeline, CancellationT
                 throw new OrchestrationException($"No handlers found for step '{currentStepName}'");
             }
 
+            sw.Restart();
```
dluc, line 187:
```suggestion
```

service/Core/Pipeline/InProcessPipelineOrchestrator.cs
```diff
@@ -180,10 +184,14 @@ public override async Task RunPipelineAsync(DataPipeline pipeline, CancellationT
...
             // Run handler
             (bool success, DataPipeline updatedPipeline) = await stepHandler
                 .InvokeAsync(pipeline, this.CancellationTokenSource.Token)
                 .ConfigureAwait(false);
+            sw.Stop();
+            this.Log.LogInformation("Pipeline '{0}/{1}' Step:{2}; Time:{3}", pipeline.Index, pipeline.DocumentId, currentStepName, sw.Elapsed);
+
```
dluc, line 194:
```suggestion
            sw.Stop();
            this.Log.LogInformation("Pipeline '{0}/{1}' Step:{2}; Time:{3}", pipeline.Index, pipeline.DocumentId, currentStepName, sw.Elapsed);

```

service/Core/Pipeline/InProcessPipelineOrchestrator.cs
```diff
@@ -2,6 +2,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
```
dluc, line 5:
```suggestion
```

dluc:
PR updated. If the code is still working it could be merged as is. Handlers now can be configured in the service without touching dependency injection and other files (see appsettings.json list of handlers)

service/Core/Handlers/ParallelGenerateEmbeddingsHandler.cs
```diff
@@ -0,0 +1,171 @@
...
+    public async Task<(bool success, DataPipeline updatedPipeline)> InvokeAsync(
+        DataPipeline pipeline, CancellationToken cancellationToken = default)
+    {
+        this._log.LogDebug("Generating embeddings, pipeline '{0}/{1}'", pipeline.Index, pipeline.DocumentId);
+
+        foreach (var uploadedFile in pipeline.Files)
+        {
+            // Track new files being generated (cannot edit originalFile.GeneratedFiles while looping it)
+            Dictionary<string, DataPipeline.GeneratedFileDetails> newFiles = new();
+            var throttler = new SemaphoreSlim(initialCount: Environment.ProcessorCount);
```
dluc, line 65: this should be disposed after the tasks are done

service/Core/Handlers/ParallelSummarizationHandler.cs
```diff
@@ -0,0 +1,215 @@
...
+    public async Task<(bool success, DataPipeline updatedPipeline)> InvokeAsync(
+        DataPipeline pipeline, CancellationToken cancellationToken = default)
+    {
+        this._log.LogDebug("Generating summary, pipeline '{0}/{1}'", pipeline.Index, pipeline.DocumentId);
+
+        foreach (DataPipeline.FileDetails uploadedFile in pipeline.Files)
+        {
+            // Track new files being generated (cannot edit originalFile.GeneratedFiles while looping it)
+            Dictionary<string, DataPipeline.GeneratedFileDetails> summaryFiles = new();
+            var throttler = new SemaphoreSlim(initialCount: Environment.ProcessorCount);
```
dluc, line 76: this should be disposed after the tasks are done

KSemenenko:
I will check this code

KSemenenko:
@dluc  I made some changes in the processing, what do you think about it? I now prefer a more standard way - parallel foreach. Also, I think it can be just part of a regular Handler, as it really relies on asynchronous operations. As an option, we could consider how many threads this should be distributed over, because I feel that supporting 2-3 handlers might be challenging.

KSemenenko:
also I use `lock (summaryFiles)` becase it's much faster then ConcurrentQueue or so.
and this is private class, so shoud'be be any side effects.

dluc (approved):
I added some tests and renamed the handlers, not to replace the default ones. From my tests the parallel handler shows a faster execution, while the summarization handler takes about the same time to generate a summary. The handlers can be used on demand, while the default ones are still in use.

KSemenenko:
This is amazing! Thanks a lot!

PR #155, opened 2023-11-14
Changed implementation of current EmbeddedPrompt to IPromptSupplier w…

### LINKED ISSUE #110: [QUESTION] Is there a way to substitute default answer_with_facts.txt prompt
Hey guys!

I'm wondering if there is a way to substitute default Completion API prompt to my custom one? Or maybe I can somehow inject my custom SearchClient that would also solve my issue.

Kind regards,
Dawid

lecramr:

Allow to customize the prompts used for QnA and summarization.

Resolves #110 

* Add a new interface IPromptProvider, by default implemented by the existing EmbeddedPrompt class (renamed to `EmbeddedPromptProvider`)
* Add new memory builder methods to allow overriding the default prompt provider.


lecramr:

dluc:
@lecramr thanks for the PR! I made a few changes mostly in the memory builder, and added an example. Let me know if this is still in line with your idea - LGTM

lecramr:
Great, I have looked at your changes and like them very much! Many thanks for the quick feedback and for the revision.

PR #156, opened 2023-11-15
Fix MsPowerPointDecoder hidden slide detection

anthonypuppo:
Fixes an issue where previously hidden/unhidden slides would be incorrectly skipped as the `Show` property would now have a non-null value of `true`.

Require the slide `Show` property to be explicitly false when determining if it should be skipped due to visibility.


dotnet/CoreLib/DataFormats/Office/MsPowerPointDecoder.cs
```diff
@@ -96,9 +96,7 @@ public class MsPowerPointDecoder
 #pragma warning restore CA1508
                 {
                     // Check if the slide is hidden and whether to skip it
-#pragma warning disable CS8625 // the property is null when the slide is visible
-                    if (skipHiddenSlides && slidePart.Slide.Show! != null) { continue; }
-#pragma warning restore CS8625
+                    if (skipHiddenSlides && ((bool)(slidePart.Slide.Show ?? true)) is false) { continue; }
```
dluc, line 99: this is odd, I tested the previous code with the pptx included in the examples. Are there multiple variations of how the Show prop is set?

anthonypuppo: Yup. Threw me for a loop when a slide on a presentation I had was being constantly skipped. Couple searches led me to the MS docs and I realized I had previously toggled the visibility of that slide...

> `If you requested that the code should limit the return value to include only visible slides, the code must filter its collection of slides to include only those slides that have a Show property that contains a value, and the value is true. If the Show property is null, that also indicates that the slide is visible. This is the most likely scenario — PowerPoint does not set the value of this property, in general, unless the slide is to be hidden. The only way the Show property would exist and have a value of true would be if you had hidden and then unhidden the slide.`

Therefore `Show` can exist in three different states, just checking for `null` isn't enough.
- `null` default, visibility has never been altered
- `false` hidden
- `true` previously hidden but now shown

Docs: https://learn.microsoft.com/en-us/office/open-xml/presentation/how-to-retrieve-the-number-of-slides-in-a-presentation-document?tabs=cs-0%2Ccs-1%2Ccs-2%2Ccs-3%2Ccs-4%2Ccs#retrieving-the-count-of-visible-slides

dlucr: thanks for the info, could you include all that information as a comment in the code?

on a different note, the syntax is quite hard to read...

```csharp
if (skipHiddenSlides && ((bool)(slidePart.Slide.Show ?? true)) is false) { continue; }
```

What about something more explicit like

```csharp
bool isVisible = !slidePart.Slide.Show.HasValue || slidePart.Slide.Show;
if (skipHiddenSlides && !isVisible) { continue; }
```

anthonypuppo: Updated. Since it can be `null` the easiest way to avoid a compiler warning is to coalesce the value. I can change it to something more along the lines of what you proposed if it's still unclear.

```csharp
// Check if the slide is hidden and whether to skip it
// PowerPoint does not set the value of this property, in general, unless the slide is to be hidden
// The only way the Show property would exist and have a value of true would be if the slide had been hidden and then unhidden
// - Show is null: default, slide is visible
// - Show is false: the slide is hidden
// - Show is true: the slide is visible
bool isVisible = slidePart.Slide.Show ?? true;
if (skipHiddenSlides && !isVisible) { continue; }
```

dluc (approved):

PR #163, opened 2023-11-17
Replace GPT3Tokenizer (p50k_base) with Microsoft.DeepDev.Tokenizer (cl100k_base)

anthonypuppo:
The existing tokenization logic is for the legacy GPT-3 and Codex models. The implementation should be updated to support the newer models (GPT-3.5, GPT-4, etc) for more precise token counts.

Remove the existing tokenization logic. Add reference to `Microsoft.DeepDev.Tokenizer` and embed `cl100k_base.tiktoken` in the assembly. Update code references as necessary.


dluc:
Might need a re-do, because we've made several changes. The solution now supports multiple tokenizers, allowing to choose the right one when selecting a model. I would leave the existing tokenizer, and simply add the new one as another option, then users can pick up the right one.

anthonypuppo:
Closing due to how stale this branch has gotten.

PR #170, opened 2023-11-23
Update Document.cs

KiisMyGun:
Check fileName error
Add message to coder

This my code  ：
```
documentName = "1700736918251108_VEO_Botox study safety report (interim).docx";

var result = await _semanticMemoryClient.ImportDocumentAsync(
    stream,
    documentName,
    documentId, 
    null, 
    _config.Services.Qdrant.DefaultIndex);
```

When the filename does not comply with the rules, an error is reported
---System.ArgumentOutOfRangeException: The document ID contains invalid symbols (Parameter 'id')
This prompt misleads me to spend time troubleshooting ID issues

### DIFF
--- a/service/Abstractions/Models/Document.cs
+++ b/service/Abstractions/Models/Document.cs
@@ -127,7 +127,7 @@ public class Document
 
         if (!IsValid(id))
         {
-            throw new ArgumentOutOfRangeException(nameof(id), "The document ID contains invalid chars (allowed: A-B, a-b, 0-9, '.', '_', '-')");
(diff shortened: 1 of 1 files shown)


dluc:
@KiisMyGun in your example, what is the value of `documentId`?

I've tried the following code, and everything worked with no exceptions

```csharp
MemoryStream data = new("hello world"u8.ToArray());

string documentId = Guid.NewGuid().ToString();

string documentName = "1700736918251108_VEO_Botox study safety report (interim).txt";

await memory.ImportDocumentAsync(
    content: data,
    fileName: documentName,
    documentId: documentId,
    tags: null,
    index: "tests",
    steps: null
);
```

KiisMyGun:
@dluc Yes,There is no such issue now。
Perhaps there is an issue with the package I previously introduced（Microsoft.SemanticMemory.Core(0.3.231009.6-preview)）

PR #171, opened 2023-11-26
[SearchClient, QueryClient] Add interface segregation & custom regist…

phykas:
Allow to customize the search client and query client used for QnA and summarization. Useful for testing different scenarios, allows for custom documents ranking algorithms, custom querying pre-processor etc.

- Add a 2 new interfaces ISearchClient & IQueryClient, by default implemented by the existing SearchClient class
- Add new memory builder methods to allow overriding the default search client & query client provider.
- Added example project with MyCustomInMemorySearchClient to show how to use it


phykas:

phykas:
@dluc Thanks for your answer, indeed this introduces sufficient flexibility and SearchClientConfiguration looks good.

Issue #107, opened 2023-10-19
[QUESTION] Should temp files be deleted after ImportDocumentAsync completed?


luismanez:
I've been doing some tests with ```memory.ImportDocumentAsync``` (either using Local file system and Azure Blob storage). In both cases, the temp files are not deleted when the pipeline is successfully finished. 
As they are temp files, shouldn't be deleted by default after pipeline complete?

According to the source code, the ```BaseOrchestrator``` class has a ``CleanUpAfterCompletionAsync``` method, but that method is only executed if you add the _delete_document_ handler to the pipeline. Besides, the code in the orchestrators leads to misunderstanding:

```csharp
if (pipeline.Complete)
        {
            this.Log.LogInformation("Pipeline '{0}/{1}' complete", pipeline.Index, pipeline.DocumentId);

            // Save the pipeline status. If this fails, the system should retry the current step.
            await this.UpdatePipelineStatusAsync(pipeline, cancellationToken).ConfigureAwait(false);

            await this.CleanUpAfterCompletionAsync(pipeline, cancellationToken).ConfigureAwait(false);
        }
```

It says that when the pipeline is complete, it will do a CleanUp, which is exactly what I'd want and expect. However, as said before, that method only runs when the DeleteDocument or DeleteIndex handlers are in the pipeline:

```csharp
if (pipeline.IsDocumentDeletionPipeline())
// stuff

if (pipeline.IsIndexDeletionPipeline())
// stuff
```

Is there any way to really CleanUp those temp files?

Thanks!

dluc:
hi @luismanez the files in storage, currently, are not considered "temp files". The plan (still under development) is to allow users to open the relevant files when receiving an "answer". The "citations" coming back with an answers will allow to reach and download/open the files stored. This is still TBD, e.g. it should be configurable, and we have to consider the impact of duplicating files. Depending on how one uses the service, the files could be considered temporary, an unwanted duplication, or useful.

About the "CleanUp" step, that has a different purpose, and it's mostly internal, I would suggest ignoring it and not relying on that. Its goal is keeping the storage consistent when a deletion is requested (in your case you're not asking for a deletion, but uploading information).

Considering we'd like to leave this open and configurable, please feel free to say how you would like to configure this and how you're using citations (if at all), so we can design accordingly :-)

Thanks!

luismanez:
[...]

PR #177, opened 2023-11-30
New handler added to allow the deletion of the generated files.

luismanez:
Added a new Handler that allows the deletion of the generated "physical" (disk, blob...) files after a document has been imported.

This should address this question: https://github.com/microsoft/kernel-memory/issues/107

This new handler as not been added to any default pipeline, so nothing changes for current clients. However, the Handler is not present, so it can be passed as _step_ when calling any of the current ```ImportDocumentAsync````
i.e:
```csharp
await memory.ImportDocumentAsync(
    "sample-Wikipedia-Moon.txt",
    steps: new[] { Constants.DeleteDocumentGeneratedFilesPipelineStepName });
```

@dluc this solution would give us the functionality we need, as would avoid any compliance issue with having enterprise docs out of M365. It's not a perfect solution, and I'm happy to make any changes if providing guidance. 
Note: There's still the pipeline_status.json file, but that's ok for us, and can be addressed with a container policy.

Thanks.

### DIFF
--- a/service/Abstractions/Constants.cs
+++ b/service/Abstractions/Constants.cs
@@ -57,16 +57,26 @@ public static class Constants
     public const string HttpIndexPlaceholder = "{index}";
     public const string HttpDocumentIdPlaceholder = "{documentId}";
 
-    // Handlers
-    public const string DeleteDocumentPipelineStepName = "private_delete_document";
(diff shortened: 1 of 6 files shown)


dluc:
looking into it, stay tuned :-)

dluc (approved):

PR #183, opened 2023-12-03
Using Semantic Kernel ITextCompletion As ITextGeneration

xbotter:

This seems somewhat recursive, to use Semantic Kernel as the infrastructure for Kernel Memory. However, the interfaces of both are very similar. So, I added an adapter that can directly use SK's ITextCompletion to maximize the reuse of LLM adapted for SK.


dluc:
could you upgrade the code to work with SK RC2 please?

xbotter:
Updated

examples/105-dotnet-SemanticKernel-TextCompletion/Program.cs
```diff
@@ -0,0 +1,34 @@
...
+var embeddingDeployment = Env.Var("AOAI_DEPLOYMENT_EMBEDDING");
+
+var memory = new KernelMemoryBuilder()
+    .WithSemanticKernelTextGenerationService(new AzureOpenAIChatCompletionService(chatDeployment, "", endpoint, apiKey))
+    .WithCustomEmbeddingGeneration(new AzureOpenAITextEmbeddingGeneration(embeddingDeployment, "", endpoint, apiKey))
+    .Build<MemoryServerless>();
+
+// using document form 203-dotnet-using-core-nuget
+
+await memory.ImportDocumentAsync("sample-SK-Readme.pdf", documentId: "doc001");
```
dluc, line 19: I'm getting `Microsoft.KernelMemory.KernelMemoryException: File not found: 'sample-SK-Readme.pdf'` here (this project doesn't see artifacts from example 203, why not include the PDF?)

dluc (changes requested):
I've made a few changes, mostly moving the files from Core to Abstractions. There's an error in the new example, easy fix

dluc:
putting on hold this, until https://github.com/microsoft/kernel-memory/pull/189 is merged. Some important changes to accommodate third-party/OSS models.

xbotter:
The new changes have brought some configurable options, and I will adapt the interface as soon as possible.

xbotter:
@dluc updated, please review again.

dluc:
Looks like there are more breaking changes coming from RC4, let's wait until next week to reduce churn

dluc (approved):

PR #185, opened 2023-12-05
Upgrade to Semantic Kernel RC2

dluc:

SK getting close to v1, with several major changes.

* Upgrade dependencies, code, syntax, etc.
* Bump KM version number.
* Minor upgrade also Azure.Search.Documents and other dependencies.


service/Abstractions/SKExtensions/KernelFunctionExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        var args = new KernelArguments();
+        if (function.Metadata.Parameters.Count >= 1)
+        {
+            // Native functions
+            args[function.Metadata.Parameters[0].Name] = text;
+        }
+        else
+        {
+            // Semantic function, this works as long as they follow
```
dluc, line 38:
```suggestion
            // Semantic functions. This works as long as they follow
```

PR #189, opened 2023-12-07
Better support for token count and tokenization depending on the model in use

dluc:

Allow to use different models with different tokenization characteristics.

* **New interfaces** required for text and embedding generation. 
* Handlers, TextChunker and SearchClient use the **current model** to check token count, to check partition size, to check summarization settings, etc.
* Text generation backends must provide a MaxToken (configurable) in order to check for correct prompt sizes
* Embedding generation backends must provide a MaxToken (configurable) in order to check for correct partitioning/chunking
* All LLMs must provide a tokenizer, in order to count the token flowing through. The system falls back to the default GPT3 tokenizer if none is provided
* Embedding and Text model details can be configured in appsettings.json or passed with the usual "With*" builder methods


service/Core/DataFormats/Text/TextChunker.cs
```diff
@@ -243,7 +244,8 @@ private static (List<string>, bool) Split(ReadOnlySpan<char> input, string? inpu
         List<string> result = new();
         var inputWasSplit = false;
 
-        int inputTokenCount = tokenCounter is null ? GetDefaultTokenCount(input.Length) : tokenCounter(inputString ??= input.ToString());
+        // int inputTokenCount = tokenCounter is null ? GetDefaultTokenCount(input.Length) : tokenCounter(inputString ??= input.ToString());
```
dluc, line 247:
```suggestion
```

PR #197, opened 2023-12-12
Tag synthetic data and allow to retrieve it

dluc:

Similarly to the summarization handler, ingestion pipelines can be customized to generate and store synthetic data. However, there's no simple mechanism to retrieve only the synthetic memory, in particular when documents chunks are generated too.

* Change pipeline status format allowing to have different tags on generated files.
* Change SearchClient to allow empty queries when using filters.
* New reserved tag usable to store the type of synthetic data, e.g. "summary". See SummarizationHandler to reuse the same in your custom handlers.
* Memory API extension methods making it easier to retrieve synthetic data like summaries.
* Add example 106 showing how to generate and retrieve summaries

How to generate summaries:
```csharp
await memory.ImportDocumentAsync(new Document("doc1")
        .AddFile("file4-SK-Readme.pdf")
        .AddFile("file5-NASA-news.pdf"),
    steps: Constants.PipelineOnlySummary);
```

How to retrieve summaries:
```csharp
var results = await memory.SearchSummariesAsync(filter: MemoryFilters.ByDocument("doc1"));
```

How to retrieve any synthetic data:
```csharp
var results = await memory.SearchSyntheticsAsync("summary", filter: MemoryFilters.ByDocument("doc1"));
```


dcakirkaya:
Could you explain the mechanism to retrieve only synthetic memories in the previous version (without these changes) ?

"there's **no simple mechanism** to retrieve only the synthetic memory, in particular when documents chunks are generated too"

(unfortunately I can't upgrade to the new version yet, I was desperately looking for a way to access the summaries generated during ImportDocumentAsync but couldn't manage to get them).

PR #204, opened 2023-12-14
Added search by id in AzureAISearchMemory

JohanYman:

I need the ability to conduct searches by ID within the search index. The content is divided into chunks, and I need a method to retrieve the chunks preceding and following the one identified by the vector search.


JohanYman:
@microsoft-github-policy-service agree company="Visma Spcs AB"

dluc:
unfortunately we cannot take the PR as it is, because the concept would work only for Azure AI Search, leaving out the other storage types, forking the memory API in two.

So this is the scenario:

> I need a method to retrieve the chunks preceding and following the one identified by the vector search.

A couple of questions:

1. Vector search always returns more than 1 record, would you want the "before and after" for each of them? How many?

2. If a document is split into 1000 chunks, and vector search returns, **in this precise order**, chunks 85, 411, 420, 86, 7, 6 - what would you expect to receive back? Note:
   * 86 is close to 85, but is less relevant than 411 and 420.
   * both 411 and 420 are relevant but not what is in the middle
   * 7 and 6 are relevant, but 7 is more relevant than 6

3. What's the use case for the chunks before and after? They are not "relevant" to the search, so how do you plan to use them?

Thanks!

dluc:
@JohanYman have you considered using the existing `IMemoryDb.GetListAsync` method?

```csharp
records = memoryDb.GetListAsync(index,
    filters: new [] { MemoryFilters.ByDocument(docId) }, 
    limit: int.MaxValue)
```

JohanYman:
I have tried to use MemoryFilters.ByDocument(docId) with less success

Our system holds documents broken into sections. Each section has a unique identifier like "DocumentName_ChunkIndex." When a user asks a question, we find the top three related sections using GetSimilarListAsync. To provide a complete answer, we also retrieve the sections before and after the match within the same document, as long as they exist. 
In our experimentation this provides invaluable context when answering user questions in a RAG fashion, where an adjacent chunk is very much related but not semantically enough to garner a high enough search hit.

Simplified Example: Imagine we have a document split into parts and a user asks about "Renewable Energy." Our search returns section "EnergyReport_02" as a top match.
We then automatically fetch "EnergyReport_01" and "EnergyReport_03."
This expanded content from sections 01 to 03 is fed into the system to generate a well-informed answer.

JohanYman:
@dluc  I have implemented this in the interface [IMemoryDb.cs](https://github.com/microsoft/kernel-memory/pull/204/files#diff-067d6fe9026c4d5a6e5d763edfb66d179c5772a24a4fc3faa78b1fb0aa7e99da) and the implemented classes as well, and added a test.

dluc:
> @JohanYman have you considered using the existing IMemoryDb.GetListAsync method?

```csharp
records = memoryDb.GetListAsync(index,
    filters: new [] { MemoryFilters.ByDocument(docId) }, 
    limit: int.MaxValue)
```

> @dluc I have tried to use MemoryFilters.ByDocument(docId) with less success

`GetListAsync` returns a list of chunks, as you describe, and given the filter, it returns chunks that are part of the document. I'd like to understand what is missing or why this list doesn't contain what you are looking for.

JohanYman:
if you add

dluc:
> `Assert.Equal() Failure: Values differ Expected: 1 Actual: 0`

that's because a `MemoryRecord` is not a `Document`, and MemoryRecord.Id is different from Document.Id. Try this:

```csharp
var simpleVecDb = new SimpleVectorDb(new SimpleVectorDbConfig(), new FakeEmbeddingGenerator());

var memory = new KernelMemoryBuilder()
    .WithCustomMemoryDb(simpleVecDb)
    .WithOpenAIDefaults("sk-.........")
    .Build<MemoryServerless>();

// Two ways to have the document ID, either it's generated by the memory...
var documentId1 = await memory.ImportWebPageAsync(
    "https://raw.githubusercontent.com/microsoft/kernel-memory/main/README.md",
    index: "tests");

// ... or you can set a Document Id when loading data
var documentId2 = "myId2";
await memory.ImportWebPageAsync(
    "https://raw.githubusercontent.com/microsoft/kernel-memory/main/README.md",
    index: "tests",
    documentId: documentId2);

// ==== FETCH ALL CHUNKS =====

var records1 = await simpleVecDb.GetListAsync("tests",
    filters: new[] { MemoryFilters.ByDocument(documentId1) },
    limit: int.MaxValue).ToArrayAsync();

Console.WriteLine(records1.Length); // ==> 7 chunks

var records2 = await simpleVecDb.GetListAsync("tests", 
    filters: new[] { MemoryFilters.ByDocument(documentId2) },
    limit: int.MaxValue).ToArrayAsync();

Console.WriteLine(records2.Length); // ==> 7 chunks
```

JohanYman:
I think the `GetByDocumentIdAsync` function could be a more efficient alternative to the current GetListAsync method. It appears to enhance efficiency by targeting specific memory areas associated with a searched ID rather than using the broader approach of GetListAsync. This could lead to better performance through reduced computational overhead and less complex post-processing. It seems well-suited for document retrieval in the vector search context, with potential benefits for system scalability and processing. The function is also set to be renamed to `GetByMemoryRecordIdAsync`, seemingly for greater clarity.

JohanYman:
> @microsoft-github-policy-service agree company="Visma Spcs AB"

jimmystridh:
@microsoft-github-policy-service agree company="Visma Spcs AB"

JohanYman:
@microsoft-github-policy-service agree company="Visma Spcs AB"
Not sure what to add more here @dluc

jimmystridh:
Hi @dluc, have you had a chance to look at this?

dluc:
> Hi @dluc, have you had a chance to look at this?

hi, yes I'm still the opinion that this is doable with `MemoryFilters.ByDocument(docId)`.
These are the steps I would use in the scenario you described:

* `list1`: find records using `GetSimilarListAsync(query)`
* for each `record` in `list1`:
   * extract `document ID` and `chunk number N`
   * `list2`: load all document records for `document ID` using `GetListAsync(filters: new [] { MemoryFilters.ByDocument(docId) }, limit: int.MaxValue)`
    * keep only `M records` before and after `chunk number N`

JohanYman:
> > Hi @dluc, have you had a chance to look at this?
> 
> hi, yes I'm still the opinion that this is doable with `MemoryFilters.ByDocument(docId)`. These are the steps I would use in the scenario you described:
> 
> * `list1`: find records using `GetSimilarListAsync(query)`
> * for each `record` in `list1`:
>   
>   * extract `document ID` and `chunk number N`
>   * `list2`: load all document records for `document ID` using `GetListAsync(filters: new [] { MemoryFilters.ByDocument(docId) }, limit: int.MaxValue)`
>   * keep only `M records` before and after `chunk number N`

Loading all documents can be inefficient because larger documents contain many sections. It appears to be significantly more efficient to only load the documents that I truly need.

dluc:
The code in the dev branch should cover this request. There's a new tag added to memory records, adding the chunk number. Given a chunk it will be possible to retrieve adjacent ones by number.

PR #211, opened 2023-12-17
Added Example for Minimal API Implementation using MemoryServerless Client

SwagataChaudhuri:
- Implementation for .NET Core Minimal Api as a sample
- Sample can be leveraged for ingestion of documents and asking questions to the same as an API integration (for apps which do not want to deploy the Memory Service but want to leverage API for integration)
 
- Implementation of document ingestion and query using ```MemoryServerless``` Client
- Exposes endpoint for ingestion of multiple files in a document
- Exposes endpoint for querying the ingested data
- Example can be leveraged by users who need to leverage an api integration to interact with ```KernelMemory``` without deploying the ```MemoryService```


SwagataChaudhuri:
@dluc, Hope you are doing good !! Can you please help in reviewing this PR? Please let me know in case you have any suggestions would be happy to contribute here

dluc:
@SwagataChaudhuri thanks for the PR. I was wondering, should we extend the existing minimal API adding those extra features like CORS? I'm not sure about having an example of a web API when the repo includes and existing web API, considering that we offer also a web client, and both provides the same entire memory API, vs a new endpoint. What do you think?

SwagataChaudhuri:
@dluc , completely agree with you but a small concern I see with web service is that since that needs to be run from the source as of now and I thought of creating a sample using the kernel memory package so that developers can look up to that as a sample and can create their own.

dluc:
> @dluc , completely agree with you but a small concern I see with web service is that since that needs to be run from the source as of now and I thought of creating a sample using the kernel memory package so that developers can look up to that as a sample and can create their own.

that's correct, though we're about to change that, the Service project will be using the packages like other apps. So in essence, one still needs to download the Service code to run it (like one would download the example here) but everything else will come via nuget.  

My main concern though, is that the Service project has a minimal API designed so that one doesn't have to write any custom HTTP client, leveraging https://www.nuget.org/packages/Microsoft.KernelMemory.WebClient, which has the same methods of the Serverless memory, for instance:

```csharp
// Instantiate HTTP client talking to KM service
var memory = new MemoryWebClient("http://127.0.0.1:9001/", "secret api key");

await memory.ImportDocumentAsync("file1.pdf");
```

SwagataChaudhuri:
This might add to a duplicate then, let me close this PR

Issue #218, opened 2023-12-22
Make the `Service` available as a NuGet package


aaronpowell:
I'm wanting to host KM as a service alongside other services that I have in an application I'm building but currently there's no easy way to do this.

The easiest way to do this is to clone the KM repo and run the service from there, but that means it's disconnected from the project that you're trying to integrate with.

An alternative solution is to copy the `Program.cs` from the Service project into your own Minimal API project and add the appropriate NuGet packages, but that's a bit risky for if/when the APIs that it consumes change.

## Proposal

Another option would be to make the part of the Service project that registers the appropriate services and creates the routes, into a NuGet package.

Then, you as the consumer of the NuGet package you would do `builder.Services.AddKernelMemory()` and use a fluent API to configure that with the AI services/memory store/etc., and then `app.MappKernelMemory()` to add the appropriate routes.

dluc:
@aaronpowell I just merged a PR changing the structure of Service.csproj, who is now depending on nugets instead of having direct references to the code. "Service" is now basically 3 files:

* **Program.cs**: minimal.API
* **ServiceConfiguration+KernelMemoryBuilderExtensions**: parse appsettings.json and create memory instance
* **HttpAuthHandler**: custom code to handle API keys

and the project
```xml
<ItemGroup>
    <PackageReference Include="Microsoft.KernelMemory.Core" Version="0.23.231224.1" />
    <PackageReference Include="Microsoft.KernelMemory.Postgres" Version="0.3.231224.1" />
</ItemGroup>
```

What about using Microsoft.KernelMemory.Core the same way that Program.cs does?  Otherwise, what would the nuget offer?

dluc:
> Another option would be to make the part of the Service project that registers the appropriate services and creates the routes, into a NuGet package.

you would be ok injecting KM minimal API, injecting KM URLs, as they are, into your web service? I think that would work

aaronpowell:
@dluc I see the changes that you've done, but I was thinking of going even further with it.

I've created #222 as a draft to start what I was thinking about. Let's discuss it more and see if we are aligned in thinking, as I'm happy to keep working away on this as I see some real use cases for it.

dluc:
Seems reasonable, I need to dig deeper into the code and how that might impact development. As you can see from https://github.com/microsoft/kernel-memory/pull/223 we're moving all extensions to packages, and Service to be a consumer of these, so I need to evaluate the change when all of that is in place. Thanks!

[...]

PR #222, opened 2023-12-28
NuGet packaging for Service logic

aaronpowell:

Addresses #218 

This moved the core logic of the `Service` project out to a new project, `Service.Core`, which can be shipped as a NuGet package that anyone could add to their own project.

I've then retrofitted the current `Service` project to use `Service.Core` as an example.

There are two main phases to the setup, adding Kernel Memory to the `appBuilder` and then adding the HTTP endpoints to the `webapp`:

- `appBuilder.AddKernelMemoryWithDefaults();`
  - This will use the defaults from the config file, otherwise you can use `appBuilder.AddKernelMemory(kernelMemoryBuilder => { ... });` and in the callback you can do your own manual wiring up of the services and what not for Kernel Memory
- `app.AddKernelMemoryEndpoints();`
  - ~Adds the HTTP endpoints that Kernel Memory supports, plus wiring them up with OpenAPI and the auth header~

### Next steps

For easier integration into an ASP.NET Core app there's a few other things that would need to be done:

- ~Route prefixing - currently you will register all routes at the root on the host, but you may want to provide a prefix, so the `AddKernelMemoryEndpoints` should support that~ Implemented are argument to `AddKernelMemoryEndpoints`
- ~Flexibility on API auth - currently it supports hand-rolled token auth, but if you want to secure with EntraID, OIDC, etc., then there needs to be a way to provide your auth rather than use the default~ Removed the auth mapping and instead surfacing that up to be attached by the host (and refactored `Service` project to do it that way)
- ~Remove OpenAPI registration - this should be left back to the owner ASP.NET Core app and not driven by config~ Done


aaronpowell:
@dluc this is ready for review and feedback

dluc (approved):

PR #228, opened 2023-12-30
Allow passing custom HttpClient

qihangnet:
## Motivation and Context
For some businesses and their teams, there is often a need for more detailed and personalized configuration when using HttpClient to make network requests. Semantic Kernel already supports this. However, inconsistencies in operation were found when using Semantic Kernel and Kernel Memory synchronously. To address this issue, this improvement PR has been submitted, aiming to standardize the corresponding features to be on par with the design of Semantic Kernel. Reference: https://github.com/microsoft/semantic-kernel/pull/743

## Description

Introduce a new optional parameter named `httpClient`. By setting its default value to `null`, we ensure that the existing user's code remains unaffected and backward compatible.


qihangnet:

dluc (approved):

PR #237, opened 2024-01-04
Reorg Redis folders

dluc:
* Move Redis tests under /extensions/Redis
* Allow to configure Redis tests and use OpenAI embeddings


slorello89 (approved):
LGTM 👍

PR #243, opened 2024-01-05
Redis: Tests, Filters, Index lists, Upsert

slorello89:
More RedisMemory stuff

1. Adding Functional Test Project (will need an instance of Redis Running)
2. When listing indexes, prefix is stripped off the index name.
3. There were some filter semantics that weren't properly implemented (everything was using AND semantics)
4. Index existence is now checked on Upsert. Upsert now leverages a lua script that:

* Checks if the index exists using [`FT.INFO`](https://redis.io/commands/ft.info/)
* If index exists, then delete and insert the document (that way the upsert is idempotent)
* If the index does not exist, return false
* if an error occurs, return the error.

If the script returns false, create the index, and then re-run the script. In the case of concurrency this may result in multiple `FT.CREATE` calls being executed, but that should be fine.

The motivation behind this is to prevent you from needed to make multiple Round trips to Redis


extensions/Redis/Redis.FunctionalTests/Redis.FunctionalTests.csproj
```diff
@@ -0,0 +1,46 @@
...
+        <ProjectReference Include="..\Redis\Redis.csproj" />
+    </ItemGroup>
+
+    <ItemGroup>
+      <None Update="appsettings.json">
+        <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      </None>
+      <None Update="appsettings.Development.json">
+        <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      </None>
```
dluc, line 42:
```suggestion
        <None Remove="appsettings.development.json"/>
        <Content Include="appsettings.development.json" Condition="Exists('appsettings.development.json')">
            <CopyToOutputDirectory>Always</CopyToOutputDirectory>
        </Content>
```

dluc: this will make sure the build passes also if the dev settings don't exist

extensions/Redis/Redis/RedisMemory.cs
```diff
@@ -111,8 +111,12 @@ public async Task<string> UpsertAsync(string index, MemoryRecord record, Cancell
     {
         var normalizedIndexName = NormalizeIndexName(index, this._config.AppPrefix);
         var key = Key(normalizedIndexName, record.Id);
```
dluc, line 113: should we validate that `record.Id` is valid, e.g. no trailing spaces, not empty, no invalid chars that could inject LUA code, to avoid errors/code injection later when passing `key` to the LUA script?

slorello89: The wire protocol for Redis [RESP](https://github.com/redis/redis-specifications/blob/master/protocol/RESP2.md)(Redis Serialization Protocol) automatically parameterizes the arguments passed into every command, so the `EVAL`/`EVALSHA` commands get `KEYS` and `ARGV` via parameterized lists of arguments. So let's take this example (stripping out payload/embedding for readability):

```
"EVALSHA" "ae03b246bcbe556aec893e033a92632cd6c03e6c" "1" "tests-default:d=2//p=a2a9371108ec466f86d1a0a37cd64750" "tests-default" "embedding" "your_embedding" "__file_part" "3bf34b3500a64953834c16eb2e4c1fd9" "user" "flash" "payload" "your_payload"
```

What's actually being sent over the wire to redis looks like this:

```
*13\r\n
$7\r\n
EVALSHA\r\n
$40\r\n
ae03b246bcbe556aec893e033a92632cd6c03e6c\r\n
$1\r\n
$53\r\n
tests-default:d=2//p=a2a9371108ec466f86d1a0a37cd64750\r\n
$13\r\n
tests-default\r\n
$9\r\n
embedding\r\n
$14\r\n
your_embedding\r\n
$11\r\n
__file_part\r\n
$32\r\n
3bf34b3500a64953834c16eb2e4c1fd9\r\n
$4\r\n
user\r\n
$5\r\n
flash\r\n
$7\r\n
payload\r\n
$12\r\n
your_payload\r\n
```

So in general injection attacks aren't really possible if you are controlling the interaction to the command interface. There have been CVEs where it was possible to use a LUA script to do nefarious things in Redis (e.g. crash Redis or privileged access to it), but that assumes the attacker can run their own scripts.

dluc (changes requested):
@slorello89 I added all the functional tests we usually run on other storage types (it's much easier now, see `DefaultTests.cs`) and found a couple of bugs:

* **ItHandlesMissingIndexesConsistently** fails throwing an exception because the index doesn't exist. The expected behavior is for the memory db to return no data without exceptions.
* A few tests fail with "StackExchange.Redis.RedisServerException **LIMIT needs two numeric arguments**".
* 3 other tests fail, probably because the memory db doesn't return the expected records and RAG produces "INFO NOT FOUND" answer

image

slorello89:
> ItHandlesMissingIndexesConsistently fails throwing an exception because the index doesn't exist. The expected behavior is for the memory db to return no data without exceptions.

Adding handling to return an empty collection when the index doesn't exist

> A few tests fail with "StackExchange.Redis.RedisServerException LIMIT needs two numeric arguments".

Fun RediSearch error. RediSearch doesn't support negative indexing into a search result. So when the cleanup bits were passing in a LIMIT of -1 the query was failing, I've pinged the PM for RediSearch, hopefully we can get some better error handling for that in the future :), for now I've had to add a bit of logic to handle this case. RediSearch doesn't support unbounded retrieval of documents from RediSearch query, there is a server-level configuration item (`MAXSEARCHRESULTS`) that dictates how many you can pull back when can change, so for now it will grab the documents in two round trips, one probing trip to see how many documents there are and to get the max documents it can pull back, and a second to get all the documents. 

> 3 other tests fail, probably because the memory db doesn't return the expected records and RAG produces "INFO NOT FOUND" answer

Yeah, this was because the `type` and `ext` tag fields weren't configured. It's too confusing to leave this as a warning so I now have it throwing an exception if it encounters an unconfigured tag field.

extensions/Redis/Redis/RedisMemory.cs
```diff
@@ -85,7 +85,7 @@ public async Task CreateIndexAsync(string index, int vectorSize, CancellationTok
     public async Task<IEnumerable<string>> GetIndexesAsync(CancellationToken cancellationToken = default)
     {
         var result = await this._search._ListAsync().ConfigureAwait(false);
-        return result.Select(x => (string)x!);
+        return result.Select(x => ((string)x!)).Where(x => x.StartsWith($"{this._config.AppPrefix}-", StringComparison.Ordinal)).Select(x => x.Substring(this._config.AppPrefix.Length + 1));
```
dluc, line 88: thoughts about making `-` part of the prefix? 

e.g. instead of 

> AppPrefix = `km` and `x.StartsWith($"{this._config.AppPrefix}-")`

setting

> AppPrefix = `km-` and then `x.StartsWith(this._config.AppPrefix)`

(plus changing other places accordingly)

slorello89: Done, my only thing would be that without the separator being outside the prefix you'll have odd looking keys/indexes. E.g. `kmtests` if the user doesn't add the prefix themselves. I'm probably overthinking as they'd never need to actually look at the keys unless they were troubleshooting something directly in Redis.

dluc: you could change the default to "km-" in `RedisConfig` ctor, similarly to the Postgres connector

slorello89: done.

dluc (approved):

PR #263, opened 2024-01-18
Update upload-example.sh

vicperdana:
fix an issue where "=" is not accepted in KM.

Error received when running upload example.sh
`Microsoft.KernelMemory.KernelMemoryException: A tag name cannot contain the '=' char`

A simple fix for an error when running the sample


dluc (approved):
Thanks!

PR #268, opened 2024-01-20
Fix the reference to security-filters.md

slapointe:
Fix a bad web page reference in the example

was: https://raw.githubusercontent.com/microsoft/kernel-memory/main/docs/SECURITY_FILTERS.md
changed to: https://raw.githubusercontent.com/microsoft/kernel-memory/main/docs/security/security-filters.md


dluc (approved):
Thanks!

PR #302, opened 2024-02-07
Redis Missing Embedding score

### LINKED ISSUE #298: [Bug] [Redis Conector] Redis relevance result is always 0 on 0.27.240205.2
### Context / Scenario

I'm using Kernel Memory in Serverless mode, in version 0.27.240205.1. When I call the 'SearchAsync' method, it returns the relevance of the search, but in the new version '0.27.240205.2', it always returns 0 for the same search.

### What happened?

In the new version '0.27.240205.2' of the NuGet package 'Microsoft.KernelMemory.MemoryDb.Redis', the relevance of searches is not displayed, always resulting in 0. I have reverted to the previous version of the package.

v.  0.27.240205.1
image

v. 0.27.240205.2

image

### Importance

a fix would make my life easier

### Platform, Language, Versions

C#, in serverless mode.

Microsoft.KernelMemory.Core" Version="0.27.240205.2" 
Microsoft.KernelMemory.MemoryDb.Redis" Version="0.27.240205.1" 
Microsoft.SemanticKernel" Version="1.3.0" 
Microsoft.SemanticKernel.Connectors.OpenAI" Version="1.3.0" 

Procesador	12th Gen Intel(R) Core(TM) i7-1255U   1.70 GHz
RAM instalada	40,0 GB (39,7 GB usable)
Tipo de sistema	Sistema operativo de 64 bits, procesador basado en x64
Lápiz y entrada táctil	La entrada táctil o manuscrita no está disponible para esta pantalla

### Relevant log output

_No response_

slorello89:
Fixes #298, Embedding score field was left out of the `RETURN` statement added in #271 to prevent superfluous embeddings from coming back. When `withEmbeddings` is set to true this is not a concern as it just pulls the entire document in Redis back. When it's set to false we were specifying which fields we wanted back, which did not include the embedding score.

Just need to append an extra field to the list of Return fields

### DIFF
--- a/.github/_typos.toml
+++ b/.github/_typos.toml
@@ -15,6 +15,7 @@ extend-exclude = [
     "encoder.json",
     "appsettings.development.json",
     "appsettings.Development.json",
+    "AzureAISearchFilteringTest.cs"
 ]
 
 [default.extend-words]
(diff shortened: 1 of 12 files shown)


dluc:
@slorello89  there seems to be a problem also with the value returned. From my tests the cosine similarity returned by Redis is very different from the one returned by other storage engines.

dluc (approved):

PR #306, opened 2024-02-11
Docker Imagepublish

cherchyk:


tools/docker/Dockerfile
dluc, line 1: after merging your PR we'll delete the /tools/docker folder -- you can ignore these files

service/Service/.gitignore
dluc, line 1: could you move this to the main .gitignore in the root? it will be useful also for the example projects

dluc:
There's a few things to adjust before merging:
* `docker run -it --rm -v ./appsettings.Development.json:/app/appsettings.Production.json IMAGE` doesn't start the service, looks like it's not picking up the configuration
* we should keep using the same port 9001, also setting env var `ENV ASPNETCORE_HTTP_PORTS 9001` - I don't think we need to export two ports, 9001 should do, because any SSL/HTTPS is usually added by the host
* files under /app should be owned by user `km` and be readonly
* I couldn't check because the image doesn't start, please make sure the process is executed by the `km` user and not `root`

Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+    # Create user
+    #Debian: useradd --create-home --user-group $USER --shell /bin/bash && \
+    adduser -D -h /app -s /bin/sh $USER && \
+    # Allow user to access the build
+    chown -R $USER.$USER /app
+
+# Define current user
+USER $USER
+
+# Used by .NET and KM to load appsettings.Productions.json
```
dluc, line 46:
```suggestion
# Used by .NET and KM to load appsettings.Production.json
```

dluc (approved):

PR #307, opened 2024-02-12
Start adding support for Mongodb Atlas database

alkampfergit:
This is a first implementation where basic unit test passes and it was tested with
some custom code. It works with cluster from M10 upward (does not work still with shared cluster).

Still have some perplexity on how to use minRelevance because I did not checked if this is a -1 1 value (cosine distance) or whatever.

Please let me know if I need to fulfill special check for the PR to be closed. Also if you think that a concrete implementation should be placed into my repository and does not find space in the official one you can tell me and reject the PR, I'll move the code in a repository of mine.

I wanted support for mongodb Atlas.


alkampfergit:

PR #311, opened 2024-02-13
Start adding support for Mongodb Atlas database

alkampfergit:
This is a first implementation where basic unit test passes and it was tested with
some custom code. It works with cluster from M10 upward (does not work still with shared cluster).

Still have some perplexity on how to use minRelevance because I did not checked if this is a -1 1 value (cosine distance) or whatever.

This code contains also the necessary code to automatically create the Vector Indexes needed to execute queries.

Please let me know if I need to fulfill special check for the PR to be closed. Also if you think that a concrete implementation should be placed into my repository and does not find space in the official one you can tell me and reject the PR, I'll move the code in a repository of mine.

I wanted support for mongodb Atlas, I've already done a pull request but it was based on dev branch that gots deleted, so I reopened on main branch.


dluc:
hi @alkampfergit thanks for the PR! 

Couple of asks:

1. could you add a script under /tools to start Mongodb Atlas with docker (assuming that's possible). See other scripts in the folder
2. about correctness, I would suggest extending these tests to compare results:
  * https://github.com/microsoft/kernel-memory/blob/main/service/tests/Core.FunctionalTests/VectorDbComparison/TestCosineSimilarity.cs. (you can use the booleans to compare just Mongodb vs SimpleVectorDb)
  * https://github.com/microsoft/kernel-memory/blob/main/service/tests/Core.FunctionalTests/VectorDbComparison/TestMemoryFilters.cs

extensions/AtlasMongoDb/AtlasMongoDb.FunctionalTests/appsettings.json
```diff
@@ -0,0 +1,57 @@
+{
+  "Logging": {
+    "LogLevel": {
+      "Default": "Information"
+    }
+  },
+  "Services": {
+    "MongoDb": {
+      "ConnectionString": "",
```
dluc, line 9: could you add a comment with an example connection string?

alkampfergit:
> hi @alkampfergit thanks for the PR!
> 
> Couple of asks:
> 
> 1. could you add a script under /tools to start Mongodb Atlas with docker (assuming that's possible). See other scripts in the folder
> 2. about correctness, I would suggest extending these tests to compare results:
> 
> * https://github.com/microsoft/kernel-memory/blob/main/service/tests/Core.FunctionalTests/VectorDbComparison/TestCosineSimilarity.cs. (you can use the booleans to compare just Mongodb vs SimpleVectorDb)
> * https://github.com/microsoft/kernel-memory/blob/main/service/tests/Core.FunctionalTests/VectorDbComparison/TestMemoryFilters.cs

Thanks a lot for the instructions, I'll proceed to implement these tests to verify that everything is ok. For what regard docker I'll investigate: Atlas is a cloud only product, but it has a commandline that can start a local cluster for tests, but I really didn't tried vector capabilities. 

I'll proceed to check and I'll update the PR accordingly.

Thanks!

alkampfergit:
Hi @dluc 

I've implement those two tests you suggeested in the previous comment. Sadly enough MongoDb perform a normalization of the base Cosine Similarity so I needed to recalcolate in memory the correct value for the test to pass. 

I've added a Markdown file explaining how to run a local Atlas MongoDB installation locally thanks to Atlas CLI. It uses podman under the hood, but it cannot work with docker. This is th easiest way to have a local running instance of AtlasDB to perform tests.

I'm waiting for comments in the PR if Any.

Thanks.

extensions/AtlasMongoDb/AtlasMongoDb/MongoDbVectorMemory.cs
```diff
@@ -0,0 +1,472 @@
...
+        {
+            yield return FromBsonDocument(document, withEmbeddings);
+        }
+    }
+
+    public async IAsyncEnumerable<(MemoryRecord, double)> GetSimilarListAsync(
+        string index,
+        string text,
+        ICollection<MemoryFilter>? filters = null,
+        double minRelevance = 0,
```
dluc, line 164: looks like this is never used, does the method filter out records by relevance if a user passes a value greater than 0?

alkampfergit: We have a problem here, I can try to implement this but atlas does not seems to filter for cosine similarity value, because it internally normalize the score ...  The only quick way that I can do is filtering client side, stopping the enumeration when the value falls below the relevance.

Also many vectror db like pinecone and also atlas mongodb can compute similiaryt with standard euclidean distance or cosine similarity or dot product (in case of unary vectors) so I'm not sure how to interpret that minRelevance score. If you ensure me that it is always pure cosine similiarity I can easily filter out during enumeration. Thanks

dluc: Thanks, I will take a look soon. About similarity, currently the expectation is to always use Cosine Similarity and to have the same value across all the engines. There was some thought to include other metrics in future, via configuration, but the default should be the same, so users can switch storage without having having to think about the math :-)

alkampfergit: I completely agree, sadly mongodb does some normalization, but in the end you can still filter in memory. 😃

alkampfergit: Also I recalculate pure cosine similarity in memory after query, from this point of view the client will have consistent numbers.

tools/run-atlas.sh
dluc, line 1: nit: please rename to `run-mongodb-atlas.sh` and set the perms to 0750 (executable)

service/tests/Core.FunctionalTests/DefaultTestCases/DocumentUploadTest.cs
```diff
@@ -16,10 +16,18 @@ public static async Task ItUploadsPDFDocsAndDeletes(IKernelMemory memory, Action
...
 
+        DateTime startWait = DateTime.UtcNow;
         while (!await memory.IsDocumentReadyAsync(documentId: Id))
         {
             log("Waiting for memory ingestion to complete...");
             await Task.Delay(TimeSpan.FromSeconds(2));
+
+            //Avoid stalling forever, whathever bug could be in the database/pipeline this test must exit
+            //after a reasonable amount of time
+            if (DateTime.UtcNow.Subtract(startWait).TotalSeconds > 120)
```
dluc, line 27: nit: pls use `DateTimeOffset` class

extensions/AtlasMongoDb/AtlasMongoDb/Helpers/AtlasSearchHelper.cs
```diff
@@ -0,0 +1,362 @@
...
+/// <summary>
+/// <para>Wrapper for ATLAS search indexes stuff</para>
+/// <para>
+/// <ul>
+/// <li>https://www.mongodb.com/docs/v7.0/reference/method/db.collection.createSearchIndex/</li>
+/// <li>Normalizer (end of the page) https://www.mongodb.com/docs/atlas/atlas-search/analyzers/</li>
+/// </ul>
+/// </para>
+/// </summary>
+public class AtlasSearchHelper
```
dluc, line 24: can we keep this class `internal` only?

extensions/AtlasMongoDb/AtlasMongoDb/MongoDbKernelMemoryStorage.cs
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    public Task CreateDocumentDirectoryAsync(string index, string documentId,
+        CancellationToken cancellationToken = new CancellationToken())
+    {
+        //no need to create anything for the document
+        return Task.CompletedTask;
+    }
+
+    private static async Task<IAsyncCursor<GridFSFileInfo<string>>> GetFromBucketByIdAsync(string id, GridFSBucket<string> bucket, CancellationToken cancellationToken)
```
dluc, line 43: here and in other files, could you sort methods by visibility?
This is the order we try to keep across files (with few exceptions):
* public constants values / static values (possibly no public constants, they are all grouped in one file)
* private constants / static values
* public getters/setters
* ctors
* public methods
* public static methods
* private methods
* private static methods

extensions/AtlasMongoDb/AtlasMongoDb/ConfigurationHelper.cs
dluc, line 1: nit: rename file to `DependencyInjection.cs`

extensions/AtlasMongoDb/AtlasMongoDb/ConfigurationHelper.cs
```diff
@@ -0,0 +1,59 @@
...
+namespace Alkampfer.KernelMemory.AtlasMongoDb;
+
+public static partial class KernelMemoryBuilderExtensions
+{
+    /// <summary>
+    /// Adds Mongodb as a storage service
+    /// </summary>
+    /// <param name="builder">The kernel builder</param>
+    /// <param name="configuration">Configuration for Mongodb</param>
+    public static IKernelMemoryBuilder WithAtlasMemoryDb(
```
dluc, line 17: as a user I might want to use MongoDb Atlas only for memory records, keeping files on disk. Could you add more extension methods to allow that?

dluc: I would also rename the method, similar to other places. E.g. here "Atlas" I think should be "MongoDbAtlas"

dluc (changes requested):

SimonLuckenuik:
Thanks for the work @alkampfergit , any chance this could be adapted to work with Azure MongoDB for vCore?
Maybe something like a base class MongoDb, and one specialized for Atlas and one for MongodDB vCore with their specific needs?
I believe the index creation and search syntax would be different, but nothing more:
https://learn.microsoft.com/en-us/azure/cosmos-db/mongodb/vcore/vector-search#use-the-createindexes-template-to-create-a-vector-index

extensions/AtlasMongoDb/AtlasMongoDb/MongoDbKernelMemoryStorage.cs
```diff
@@ -0,0 +1,216 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.KernelMemory.ContentStorage;
+using MongoDB.Bson;
+using MongoDB.Driver;
+using MongoDB.Driver.GridFS;
+
+namespace Alkampfer.KernelMemory.AtlasMongoDb;
+
+public class MongoDbKernelMemoryStorage : MongoDbKernelMemoryBaseStorage, IContentStorage
```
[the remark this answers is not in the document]
alkampfergit: Actually MongoDb has GridFs to store files and blob. In production system I do not like to have all the blob and information on disk, also I can benefit from the ability to query from various UI interface all the record inside my mongodb for diagnostic. Clearly if you want to use only the MemoryDb part you can simply register only the MEmoryDb part completely ignoring this class.

alkampfergit:
> Thanks for the work @alkampfergit , any chance this could be adapted to work with Azure MongoDB for vCore? Maybe something like a base class MongoDb, and one specialized for Atlas and one for MongodDB vCore with their specific needs? I believe the index creation and search syntax would be different, but nothing more: https://learn.microsoft.com/en-us/azure/cosmos-db/mongodb/vcore/vector-search#use-the-createindexes-template-to-create-a-vector-index

Actually we are using MongoDbAtlas, I do not know MongodDB vCore, does it support Vector Search in the same way atlas does? (lucene index).

extensions/AtlasMongoDb/AtlasMongoDb/MongoDbKernelMemoryStorage.cs
alkampfergit: I should have fixed ordering.

extensions/AtlasMongoDb/AtlasMongoDb/Helpers/AtlasSearchHelper.cs
alkampfergit: I added internalsvisible to because it is used by test to cleanup.

extensions/AtlasMongoDb/AtlasMongoDb/AtlasMongoDb.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>net6.0</TargetFramework>
+        <RollForward>LatestMajor</RollForward>
+        <AssemblyName>Alkampfer.KernelMemory.AtlasMongoDb</AssemblyName>
+        <RootNamespace>Alkampfer.KernelMemory.AtlasMongoDb</RootNamespace>
```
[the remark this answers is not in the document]
alkampfergit: Done

tools/run-atlas.sh
alkampfergit: done

tools/create-local-atlas.md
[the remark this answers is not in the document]
alkampfergit: done

extensions/AtlasMongoDb/AtlasMongoDb/ConfigurationHelper.cs
alkampfergit: done

extensions/AtlasMongoDb/AtlasMongoDb/ConfigurationHelper.cs
alkampfergit: Done

alkampfergit:
@dluc I was pretty busy these days, but I should had resolved all the issues. Feel free to request additional modification (i've left two commits, the first is the original for the PR, the other contains all modification done as for your request)

extensions/AtlasMongoDb/AtlasMongoDb.FunctionalTests/DefaultTests.cs
```diff
@@ -0,0 +1,111 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using Alkampfer.KernelMemory.AtlasMongoDb;
+using FunctionalTests.DefaultTestCases;
+using KernelMemory.AtlasMongoDb.Helpers;
+using Microsoft.KernelMemory;
+using Microsoft.TestHelpers;
+
+namespace AtlasMongoDb.FunctionalTests;
```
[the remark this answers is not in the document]
alkampfergit: Done

SimonLuckenuik:
> > Thanks for the work @alkampfergit , any chance this could be adapted to work with Azure MongoDB for vCore? Maybe something like a base class MongoDb, and one specialized for Atlas and one for MongodDB vCore with their specific needs? I believe the index creation and search syntax would be different, but nothing more: https://learn.microsoft.com/en-us/azure/cosmos-db/mongodb/vcore/vector-search#use-the-createindexes-template-to-create-a-vector-index
> 
> Actually we are using MongoDbAtlas, I do not know MongodDB vCore, does it support Vector Search in the same way atlas does? (lucene index).

Yes it does support vector search, but the index and query syntaxes are a bit different, see here: https://learn.microsoft.com/en-us/azure/cosmos-db/mongodb/vcore/vector-search#use-the-createindexes-template-to-create-a-vector-index

Someone also started a MongoDB vCore pull request: https://github.com/microsoft/kernel-memory/pull/329

In the end, for maintenance, most probably it would be more interesting to have a common shared code base.

dviry:
Sorry for meddling, but I have been using this and I have some feedback:
- why the requirement to use Mongo v7? Most Atlas clusters (except the highest plans) cannot yet be upgraded to v7
- the [knnBeta search type is deprecated](https://www.mongodb.com/docs/atlas/atlas-search/knn-beta/), it should use the normal $vectorSearch
- there are some warnings from the NetAnalyzers regarding code-quality (#nullable)
- why hard-coding BsonDocuments everywhere instead of using the standard .NET driver? Would make it more future-proof
- the collectionName is hard coded (even the ConnectionNamePrefix cannot be changed from the default "_ix_") I would like to be able to configure that (we are using a naming convention within one Mongo db)
- I haven't seen any error-handling - should all exceptions always bubble up? In my experience the MongoDB exceptions are not always helpful and the logging might have issues with those
- it's not really a best-practice to store long strings in the Mongo _id field - this has a performance/storage impact. I also see string storage in tg___file_id[] - although it's a GUID that could be well stored in a MongoDB iuid/binary field

Just a general question - I know SemanticKernel is a different project but the connectors are very similar - couldn't we somehow share/reuse some of the codebase (or just have a wrapper?), for example [the MongoDB Connector](https://github.com/microsoft/semantic-kernel/tree/main/dotnet/src/Connectors/Connectors.Memory.MongoDB)

alkampfergit:
UPDATE: I've updated the pull request as for @dviry request, for now I've addressed support for 6 and 7 (tested with local atlas insetallation), also removed knnBeta to favor $vectorSearch.

Still the code can be improved to let the user choose naming of the collection and other minor updates.

Fixed also a problem in creation of Vector Search index where a stale vector resulted as created but was not functional.

dviry:
Thanks for considering my feedback!

One question before I update my local copy - am I missing something or is the [knnBeta](https://github.com/microsoft/kernel-memory/blob/ccf8fcfe62e509e5d85bf42cdd6322baa8fb2e23/extensions/MongoDbAtlas/MongoDbAtlas/MongoDbVectorMemory.cs#L369) still in the latest code?

Thanks

extensions/MongoDbAtlas/MongoDbAtlas/Helpers/MongoDbAtlasSearchHelper.cs
```diff
@@ -0,0 +1,303 @@
...
+    /// Delete all the indices for a specific collection
+    /// </summary>
+    /// <param name="collectionName"></param>
+    public async Task DeleteIndicesAsync(string collectionName)
+    {
+        var pipeline = new BsonDocument[]
+        {
+            new() {
+                {
+                    "$listSearchIndexes",
```
dviry, line 97: note that this only works starting M10 instances (not on "free forever" M0)

extensions/MongoDbAtlas/MongoDbAtlas/Helpers/MongoDbAtlasSearchHelper.cs
```diff
@@ -0,0 +1,303 @@
...
+        return new BsonDocument
+        {
+            { "createSearchIndexes", collectionName },
+            {
+                "indexes", new BsonArray
+                {
+                    new BsonDocument
+                    {
+                        { "name", this.GetIndexName(collectionName) },
+                        { "type", "vectorSearch" },
```
dviry, line 161: type should be ```vector```

extensions/MongoDbAtlas/MongoDbAtlas/Helpers/MongoDbAtlasSearchHelper.cs
alkampfergit: I know, do you know a way to get the list of indexes in an M0 instance? The only workaround I know is an option where you avoid the code checking and creating index and you need to create index manually.

Alternatively, I can simply try to create the index and trap exception of index already existing so it can work even in M0.

alkampfergit:
> Thanks for considering my feedback!
> 
> One question before I update my local copy - am I missing something or is the [knnBeta](https://github.com/microsoft/kernel-memory/blob/ccf8fcfe62e509e5d85bf42cdd6322baa8fb2e23/extensions/MongoDbAtlas/MongoDbAtlas/MongoDbVectorMemory.cs#L369) still in the latest code?
> 
> Thanks
> 
> EDIT: sorry my mistake - the method CreatePipeline() is still there but not referenced anywhere

Removed and amended last commit, I've left old code for review and compare but it makes sense to remove to have cleaner code.

extensions/MongoDbAtlas/MongoDbAtlas/Helpers/MongoDbAtlasSearchHelper.cs
dviry: AFAIK that won't work either, since creating SearchIndexes is also only available starting M10...

alkampfergit: The only solution would be adding an option to the configuration telling the component not to create the indexes, so when the core code of Kernel Memory tells you to create an index, you will simply skip.

dviry: I personally prefer less hidden "options" - I would rather handle that specific error (both on index check/creation) and log it accordingly. Additionally I would add it as a requirement to the connector Documentation (maybe with details on how to create it manually in case anyone wants to keep using an M0 instance for development).

alkampfergit: I'll check, I had a situation where the collection was created but the index no, this is the reason why I always check for the index. The UpdateOneAsync can be avoided, even if I thought that CreateIndexAsync() is a method that is rarely called, I'll check as soon as possible, thanks.

dluc:
For some reason the service doesn't work on my workstation, e.g. the cosine similarity test is unable to connect.

```
⚡️ ./run-mongodb-atlas.sh

[Default Settings]
Deployment Name   local6786
MongoDB Version   7.0
Port              27017

Creating your cluster local6786 [this might take several minutes]
1/3: Starting your local environment...
2/3: Downloading the MongoDB binaries to your local environment...
3/3: Creating your deployment local6786...
Deployment created!
Connection string: mongodb://root:root@localhost:27017/?directConnection=true

connection skipped
```

```
Deleting index test-cosinesimil in Microsoft.KernelMemory.MongoDbAtlas.MongoDbVectorMemory

System.TimeoutException
A timeout occurred after 10000ms selecting a server using CompositeServerSelector{ Selectors = ReadPreferenceServerSelector{ ReadPreference = { Mode : Primary } }, LatencyLimitingServerSelector{ AllowedLatencyRange = 00:00:00.0150000 }, OperationsCountServerSelector }. Client view of cluster state is { ClusterId : "1", DirectConnection : "True", Type : "Standalone", State : "Disconnected", Servers : [{ ServerId: "{ ClusterId : 1, EndPoint : "Unspecified/localhost:27017" }", EndPoint: "Unspecified/localhost:27017", ReasonChanged: "Heartbeat", State: "Disconnected", ServerVersion: , TopologyVersion: , Type: "Unknown", HeartbeatException: "MongoDB.Driver.MongoConnectionException: An exception occurred while opening a connection to the server.
 ---> System.Net.Sockets.SocketException (61): Connection refused [::1]:27017
   at System.Net.Sockets.Socket.DoConnect(EndPoint endPointSnapshot, SocketAddress socketAddress)
   at System.Net.Sockets.Socket.Connect(EndPoint remoteEP)
   at MongoDB.Driver.Core.Connections.TcpStreamFactory.Connect(Socket socket, EndPoint endPoint, CancellationToken cancellationToken)
   at MongoDB.Driver.Core.Connections.TcpStreamFactory.CreateStream(EndPoint endPoint, CancellationToken cancellationToken)
   at MongoDB.Driver.Core.Connections.BinaryConnection.OpenHelper(CancellationToken cancellationToken)
   --- End of inner exception stack trace ---
```

> "ConnectionString": "mongodb://localhost:27017/?directConnection=true&serverSelectionTimeoutMS=10000",

dluc:
I would rename these classes, to be consistent with others:

* `MongoDbVectorMemory` => `MongoDbAtlasMemoryDb`
* `MongoDbKernelMemoryStorage` => `MongoDbAtlasStorage` or `MongoDbAtlasContentStorage`

image

dviry:
As mentioned above those commands only work starting M10 Atlas instances,
not on M0 ones... You will have to create the index manually in case you
want to continue using the M0 instance.

dluc:
> As mentioned above those commands only work starting M10 Atlas instances, not on M0 ones... You will have to create the index manually in case you want to continue using the M0 instance.

I see, so we need to figure out how to fix the local docker instance, otherwise we won't be able to test the integration. Unless we get support from MongoDb and they could offer a M10 instance for test purposes.

alkampfergit:
> > As mentioned above those commands only work starting M10 Atlas instances, not on M0 ones... You will have to create the index manually in case you want to continue using the M0 instance.
> 
> I see, so we need to figure out how to fix the local docker instance, otherwise we won't be able to test the integration. Unless we get support from MongoDb and they could offer a M10 instance for test purposes. In the meantime I'll try the version available via homebrew, see if it can support the integration tests.

From what I see you are connecting in port 27017, that is the standard mongodb port, but the script to start the cluster starts in docker and expose the port in 27777 so the connection string is

mongodb://root:root@localhost:27777/?authSource=admin

I've added a comment in the file. The reason behind this is that usually, developers that works with mognodb, have a standard local instance, then when they want to test atlas, they use a different port to avoid conflict.

To simplify everything we can change the port in the sh file to use standard one.

dluc:
Update: local tests via homebrew are working.
I made some changes to functional tests and renamed the "MongoDb" config key to "MongoDbAtlas".
All functional tests seem to pass fine.

alkampfergit:
> Update: local tests via homebrew are working. I made some changes to functional tests and renamed the "MongoDb" config key to "MongoDbAtlas". All functional tests seem to pass fine.

Good, sorry for the inconvenience, I was using secret because I tested against a M20 of a customer ,so I really do not want to leak the connectino string and I forgot to change the default one to match that one of the script .sh.

extensions/MongoDbAtlas/MongoDbAtlas/MongoDbAtlasConfig.cs
```diff
@@ -0,0 +1,70 @@
...
+        return this;
+    }
+
+    public MongoDbAtlasConfig WithAfterIndexCallback(Func<Task> afterIndexCallback)
+    {
+        this.AfterIndexCallbackAsync = afterIndexCallback;
+        return this;
+    }
+
+    internal IMongoDatabase GetDatabase()
```
dluc, line 55: this method seems a bit odd, e.g. having a client inside the config class. The code and the client instance could be moved to the `MongoDbAtlasBaseStorage` class.

extensions/MongoDbAtlas/MongoDbAtlas/MongoDbAtlasMemory.cs
```diff
@@ -0,0 +1,351 @@
...
+    public async IAsyncEnumerable<MemoryRecord> GetListAsync(
+        string index,
+        ICollection<MemoryFilter>? filters = null,
+        int limit = 1,
+        bool withEmbeddings = false,
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    {
+        if (limit <= 0)
+        {
+            limit = 10;
```
dluc, line 102: this var is never used, the method should stop after returning `limit` records

extensions/MongoDbAtlas/MongoDbAtlas/MongoDbAtlasMemory.cs
```diff
@@ -0,0 +1,351 @@
...
+            .VectorSearch(m => m.Embedding, vector, limit, options)
+            .ToListAsync(cancellationToken).ConfigureAwait(false);
+
+        // If you check documentation Atlas normalize the score with formula
+        // score = (1 + cosine/dot_product(v1,v2)) / 2
+        // Thus it does not output the real cosine similarity, this is annoying so we
+        // need to recompute cosine similarity using the embeddings.
+#if DEBUG
+        //List<(double, double, double)> similarities = new();
+#endif
```
dluc, line 164: could you remove the code commented out?

Looks like there are a bunch of #DEBUG blocks that could be removed across the file

extensions/MongoDbAtlas/MongoDbAtlas/MongoDbAtlasMemory.cs
```diff
@@ -0,0 +1,351 @@
...
+        return record.Id;
+    }
+
+    private FilterDefinition<MongoDbAtlasMemoryRecord>? TranslateFilters(ICollection<MemoryFilter>? filters, string index)
+    {
+        List<FilterDefinition<MongoDbAtlasMemoryRecord>> outerFiltersArray = new();
+        foreach (var filter in filters ?? Array.Empty<MemoryFilter>())
+        {
+            var thisFilter = filter.GetFilters().ToArray();
+            var numOfFilters = thisFilter.Count(x => !string.IsNullOrEmpty(x.Value));
```
dluc, line 215: this var is never used

extensions/MongoDbAtlas/MongoDbAtlas/MongoDbAtlasMemory.cs
```diff
@@ -0,0 +1,351 @@
...
+            foreach (var singleFilter in thisFilter)
+            {
+                var condition = Builders<MongoDbAtlasMemoryRecord>.Filter.And(
+                    Builders<MongoDbAtlasMemoryRecord>.Filter.Eq("Tags.Key", singleFilter.Key),
+                    Builders<MongoDbAtlasMemoryRecord>.Filter.Eq("Tags.Values", singleFilter.Value)
+                );
+                filtersArray.Add(condition);
+            }
+
+            // All these filter if more than one must be enclosed in an end filter
```
dluc, line 226: could you revisit the comment, might be missing some word or have a typo - thanks

extensions/MongoDbAtlas/MongoDbAtlas/MongoDbAtlasStorage.cs
```diff
@@ -0,0 +1,216 @@
...
+
+    public Task CreateIndexDirectoryAsync(string index, CancellationToken cancellationToken = new CancellationToken())
+    {
+        return Task.CompletedTask;
+    }
+
+    public async Task DeleteIndexDirectoryAsync(string index, CancellationToken cancellationToken = new CancellationToken())
+    {
+        // Delete all document in GridFS that have index as metadata
+        var filter = Builders<GridFSFileInfo>.Filter.Eq("metadata.index", index);
```
dluc, line 28: var never used

dluc (changes requested):

extensions/MongoDbAtlas/MongoDbAtlas/MongoDbAtlasConfig.cs
alkampfergit: it could be done, but it is best practice to have a single Client for the entire application, so it is standard practice to leave the client in a single instance of a global class. Actually it would be also better (maybe in a subsequent change) to give the caller the ability to specify the client class from external code.


I've moved inside a dedicated class.

extensions/MongoDbAtlas/MongoDbAtlas/MongoDbAtlasMemory.cs
alkampfergit: you are right, it is a leftover for some tests I made at the beginning.

extensions/MongoDbAtlas/MongoDbAtlas/MongoDbAtlasMemory.cs
alkampfergit: I've put a check inside the loop, but this value is passed to the method that created the filter, the number of item is directly inserted in the mongodb query.

extensions/MongoDbAtlas/MongoDbAtlas/MongoDbAtlasConfig.cs
dluc: the memory db object is a singleton, that should be sufficient no?

alkampfergit:
Yes

Il dom 10 mar 2024, 19:26 Devis Lucato ***@***.***> ha
scritto:

service/tests/Core.FunctionalTests/Core.FunctionalTests.csproj
```diff
@@ -14,10 +14,8 @@
     </PropertyGroup>
 
     <ItemGroup>
```
dluc, line 16: No need to add project references here, I removed also old leftovers

extensions/MongoDbAtlas/MongoDbAtlas/MongoDbAtlasDatabaseHelper.cs
```diff
@@ -0,0 +1,26 @@
...
+namespace Microsoft.KernelMemory.MongoDbAtlas;
+
+/// <summary>
+/// Allow keeping a singleton for IMongoDatabase and MongoClient
+/// </summary>
+internal static class MongoDbAtlasDatabaseHelper
+{
+    private static MongoClient? s_client = null;
+
+    internal static IMongoDatabase GetDatabase(string connectionString, string databaseName)
```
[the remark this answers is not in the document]
alkampfergit: I changed the code, actually I've ported a code from other projects of mine, client is not static anymore but it is stored inside a static concurrent dictionary. This will ensure that we will create a single IMongoClient for each different connection string. At the same time we allow to use 2 different connections, one for storage and other for memory. 

Also used the same class from the SearchHelper class.

service/tests/Core.FunctionalTests/Core.FunctionalTests.csproj
alkampfergit: Ok you are right.

dluc (approved):

Issue #337, opened 2024-03-03
How do i set temperature when using AskAsync?


Christofferjensen:
### Context / Scenario

How do i set temperature when using AskAsync?

### Question

How do i set temperature when using AskAsync?

            var openAIConfig = new OpenAIConfig();
            new ConfigurationBuilder()
                .AddJsonFile("appsettings.json")
                .Build()
                .BindSection("KernelMemory:Services:OpenAI", openAIConfig);

            var memory = new KernelMemoryBuilder()
                .WithOpenAI(openAIConfig)
                .WithSearchClientConfig(new SearchClientConfig {
                    AnswerTokens = 400,
                    MaxMatchesCount = 10,
                })
                .WithOpenAITextGeneration(openAIConfig)
                .Build<MemoryServerless>();

            Answer = memory.AskAsync(Promt, filters: keyValuePairsColl);

Any help would be appreciated

dluc:
TopP and Temperature are hard coded in `SearchClient`. The values are set to 0 to reduce the chance of hallucinations and maximize reproducible answers. If you must change those values, there's a few options and ideas:
1. you can use KM builder's `WithCustomSearchClient` method to inject a customized version of `SearchClient`. So you could duplicate that class, maintain your own copy, with your custom settings
2. we would welcome a PR that extends the values in `SearchClientConfig` to include temperature and other settings.
3. there's a long term plan, no timeline yet, to allow passing in these values during the request. That would/will allow more use cases without the need to change configurations, redeploying the service etc. The approach needs to be designed, so this won't be available soon, so I would considering (1) or (2).

dluc:
Update: @marcominerva added new settings to `SeachClientConfig`, see https://github.com/microsoft/kernel-memory/pull/341, allowing to configure Temperature, TopP and other LLM request settings. These will soon be used by `AskAsync`, PR coming soon.

dluc:
Fixed: see #341 and #344

Issue #137, opened 2023-11-05
TextGenerationOptions is totally not used


AsakusaRinne:
`TextGenerationOptions` is a parameter of `ITextGeneration.GenerateTextAsync`. However currently it seems to be not used anywhere.

For some API service like OpenAI Chatgpt, stop sequence is not so important. However for local model inference, the model will endlessly generate response without a stop sequence.

Could you please expose `TextGenerationOptions` to `AskAsync` API to let users configure the settings themselves? It will help a lot for local LLM inference integration.

xbotter:
If possible, I hope that the method of calculating the number of tokens can also provide custom configuration.

AsakusaRinne:
Any updates? @dluc I understand that at the beginning stage of a project, there's always short of hands. Please at least let us know if it would be solved in the future.

dluc:
Sorry we didn't have an opportunity to look into this yet, but we always keep an eye on the list of open issues, so we'll provide an update as soon as possible.

AsakusaRinne:
Ok, I'm looking forward to it. Thank you for your works anyway.

dluc:
I noticed that LLama would generate tokens ad infinitum (almost, at some point it throws an exception). `SearchClientConfig.AnswerTokens` will be passed as `TextGenerationOptions.MaxTokens`.

I'll look into adding the options to the Ask API, so the behavior can be managed more easily.

AsakusaRinne:
Thank you a lot! I'm looking forward to it

dluc:
Update: @marcominerva added new settings to `SeachClientConfi`g, see https://github.com/microsoft/kernel-memory/pull/341, allowing to configure Temperature, TopP and other LLM request settings. These will soon be used by `AskAsync`, PR coming soon.

dluc:
Fixed: see #341 and #344

PR #341, opened 2024-03-05
Add properties to set SearchClient settings like Temperature and TopP

marcominerva:
With the exception of `MaxTokens`, properties of the [TextGenerationOptions](https://github.com/microsoft/kernel-memory/blob/dacc5047e339fe4073636b3d9b54c32438073978/service/Core/Search/SearchClient.cs#L341-L350) object used by `AskAsync` are hard-coded.

This PR addresses issues https://github.com/microsoft/kernel-memory/issues/137 and https://github.com/microsoft/kernel-memory/issues/337. It updates the [SearchClientConfig.cs](https://github.com/microsoft/kernel-memory/blob/main/service/Abstractions/Search/SearchClientConfig.cs) file in the **Abstractions** project adding the properties that allows to set **Temperature**, **TopP**, **PresencePenaltiy**, etc.

Once the **Abstractions** project is updated on NuGet, the **Core** project can be in turn modified to use these new settings in the [SearchClient.cs](https://github.com/microsoft/kernel-memory/blob/dacc5047e339fe4073636b3d9b54c32438073978/service/Core/Search/SearchClient.cs#L341-L350) file.

### DIFF
--- a/service/Abstractions/Search/SearchClientConfig.cs
+++ b/service/Abstractions/Search/SearchClientConfig.cs
@@ -41,40 +41,40 @@ public class SearchClientConfig
     /// </summary>
     public string EmptyAnswer { get; set; } = "INFO NOT FOUND";
 
-    //
-    // Summary:
-    //     Temperature controls the randomness of the completion. The higher the temperature,
-    //     the more random the completion.
-    public double Temperature { get; set; }
+    /// <summary>
+    /// Number between 0.0 and 2.0. It controls the randomness of the completion.
+    /// The higher the temperature, the more random the completion.
+    /// </summary>
+    public double Temperature { get; set; } = 0;
 
-    //
-    // Summary:
-    //     TopP controls the diversity of the completion. The higher the TopP, the more
-    //     diverse the completion.
-    public double TopP { get; set; }
+    /// <summary>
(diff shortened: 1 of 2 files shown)


service/Abstractions/Search/SearchClientConfig.cs
```diff
@@ -40,6 +41,42 @@ public class SearchClientConfig
     /// </summary>
     public string EmptyAnswer { get; set; } = "INFO NOT FOUND";
 
+    //
```
dluc, line 44: nit: pls use the usual documentation format (triple forward slashes, xmldocs, etc), and add valid ranges also in the comments

service/Abstractions/Search/SearchClientConfig.cs
```diff
@@ -40,6 +41,42 @@ public class SearchClientConfig
...
+    // Summary:
+    //     Temperature controls the randomness of the completion. The higher the temperature,
+    //     the more random the completion.
+    public double Temperature { get; set; }
```
dluc, line 48: nit: pls add an explicit default to 0 (same for the props below)

dluc (changes requested):

dluc (approved):

Issue #349, opened 2024-03-08
[Bug] Issue with default index name management


marcominerva:
### Context / Scenario

Qdrant config has been updated to support default index name. However, there are some points in the **Abstractions** project that statically set the index name as `Constants.DefaultIndex` if it isn't provided.

### What happened?

The **Abstractions** project contains an [IndexExtensions.cs](https://github.com/microsoft/kernel-memory/blob/main/service/Abstractions/Models/IndexExtensions.cs) file that statically sets the index name to `Constants.DefaultIndex` if no name is set:

https://github.com/microsoft/kernel-memory/blob/61fb2b5789ba51543ab947a604dc8c1db3ac8bf1/service/Abstractions/Models/IndexExtensions.cs#L5-L14

This method is called
- In the `Index `property of the [DataPipeline.cs](https://github.com/microsoft/kernel-memory/blob/main/service/Abstractions/Pipeline/DataPipeline.cs) class:
https://github.com/microsoft/kernel-memory/blob/61fb2b5789ba51543ab947a604dc8c1db3ac8bf1/service/Abstractions/Pipeline/DataPipeline.cs#L213-L217
- In the constructor of the [DocumentUploadRequest.cs](https://github.com/microsoft/kernel-memory/blob/main/service/Abstractions/Models/DocumentUploadRequest.cs) class:
https://github.com/microsoft/kernel-memory/blob/61fb2b5789ba51543ab947a604dc8c1db3ac8bf1/service/Abstractions/Models/DocumentUploadRequest.cs#L76-L78

In this way, when running for example [SaveRecordsHandler.InvokeAsync](https://github.com/microsoft/kernel-memory/blob/main/service/Core/Handlers/SaveRecordsHandler.cs#L84): https://github.com/microsoft/kernel-memory/blob/61fb2b5789ba51543ab947a604dc8c1db3ac8bf1/service/Core/Handlers/SaveRecordsHandler.cs#L84-L95
`pipeline.Index` will always have a value (if not provided, it will statically be `Costants.DefaultIndex`, as shown above). 

As a result, the [QdrantMemory.NormalizeIndexName](https://github.com/microsoft/kernel-memory/blob/main/extensions/Qdrant/Qdrant/QdrantMemory.cs#L226-L236) method:

https://github.com/microsoft/kernel-memory/blob/61fb2b5789ba51543ab947a604dc8c1db3ac8bf1/extensions/Qdrant/Qdrant/QdrantMemory.cs#L226-L236

[...]

PR #351, opened 2024-03-11
Allow to set default index name at service level

marcominerva:
This PR addresses issues https://github.com/microsoft/kernel-memory/issues/349 and allows to specify the default index name at service level.

This PR introduces a new property in the `KernelMemoryConfig` class to set the default index name, that will be always used if no one is specified, regardless the memory and content storage used.

The configuration can be done using `appsettings.json`:

```json
"KernelMemory": {
  "Service": {
    // 
  },
  "ContentStorageType": "SimpleFileStorage",
  // "AzureOpenAIText", "OpenAI" or "LlamaSharp"
  "TextGeneratorType": "",
  // Name of the default KM index/Qdrant collection, when none is specified
  "DefaultIndex": "default",
  // Data ingestion pipelines configuration.
  "DataIngestion": {
    // ...
    }
  }
```

Or directly in code:

```csharp
var kernelMemoryBuilder = new KernelMemoryBuilder()
    .With(new KernelMemoryConfig { DefaultIndex = "my_index" });
```

If the `DefaultIndex` property isn't set, the value **default** will be used.


dluc:
Thanks for sending it. There are some changes I would avoid, e.g. the pipeline definition.

I created a similar PR here https://github.com/microsoft/kernel-memory/pull/354 if you want to test it out.

dluc:
Proceeding with #354

PR #359, opened 2024-03-13
Fixed #358 - configuration on Disk from clean folder

### LINKED ISSUE #358: Running server with Simple Disk storage seems always to use volatile memory
### Context / Scenario

I configured the Server with this section for services. As you can see I uses simple queue on disk, file storage on disk and vector memory on disk. The server starts and works perfectly but all the services uses Volatile File system, not using file system on disk.

```json
 "Services": {
      "SimpleQueues": {
        "Directory": "c:\\temp\\km\\queue"
      },
      "SimpleFileStorage": {
        "Directory": "c:\\temp\\km\\storage"
      },
      "OpenAI": {
        "TextModel": "gpt-3.5-turbo-16k",
        "EmbeddingModel": "text-embedding-ada-002",
        "APIKey": "my-api-key",
        "OrgId": "",
        "MaxRetries": 10
      },
      "SimpleVectorDb": {
        "Directory": "c:\\temp\\km\\vector"
      }
    }
```

### Question

What is wrong in the configuration above? I generated with dotnet run config and it seems to be correct, but nevertheless no data is persisted on disk.

Also if I download the project in a fresh folder or delete all non source controlled file `git clean -xdf` when I run dotnet run setup I got an error because it does not find the application.development.json.

alkampfergit:

When you download the project in a fresh folder, running `dotnet run setup` in the service folder throws an error because development.json file does not exists.

Also if you configure using Disk storage (not for production but useful for testing) actually it uses volatile storage and does not persist anytying on disk.

it will fix #358

I've simply restored the generation of basic development file if not exists, also fixed configuration using "disk" storage type instead of "volatile"

### DIFF
--- a/tools/InteractiveSetup/Main.cs
+++ b/tools/InteractiveSetup/Main.cs
@@ -626,9 +626,12 @@ public static class Main
             };
         }
 
+        var queueFolderOnDisk = SetupUI.AskOpenQuestion("Directory where to store queue messages, empty to use in memory volatile queue", "", optional: true);
+
         AppSettings.Change(x => x.Services[ServiceName] = new Dictionary<string, object>
         {
-            { "Directory", SetupUI.AskOpenQuestion("Directory where to store queue messages", config["Directory"].ToString()) }
+            { "StorageType", string.IsNullOrEmpty(queueFolderOnDisk) ? "Volatile" : "Disk" },
+            { "Directory", queueFolderOnDisk }
         });
     }
 
@@ -732,9 +735,12 @@ public static class Main
             };
         }
 
+        var queueFolderOnDisk = SetupUI.AskOpenQuestion("Directory where to store files, empty to use in memory volatile storage", "", optional: true);
+
         AppSettings.Change(x => x.Services[ServiceName] = new Dictionary<string, object>
         {
-            { "Directory", SetupUI.AskOpenQuestion("Directory where to store files", config["Directory"].ToString()) }
+            { "StorageType", string.IsNullOrEmpty(queueFolderOnDisk) ? "Volatile" : "Disk" },
+            { "Directory", queueFolderOnDisk }
         });
     }
 
@@ -860,9 +866,12 @@ public static class Main
             };
         }
 
+        var queueFolderOnDisk = SetupUI.AskOpenQuestion("Directory where to store vectors, empty to use in memory volatile memory", "", optional: true);
+
         AppSettings.Change(x => x.Services[ServiceName] = new Dictionary<string, object>
         {
-            { "Directory", SetupUI.AskOpenQuestion("Directory where to store vectors", config["Directory"].ToString()) }
+            { "StorageType", string.IsNullOrEmpty(queueFolderOnDisk) ? "Volatile" : "Disk" },
+            { "Directory", queueFolderOnDisk }
         });
     }
 
--- a/tools/InteractiveSetup/Main.cs
+++ b/tools/InteractiveSetup/Main.cs
@@ -732,8 +732,6 @@ public static class Main
             };
         }
 
-        var queueFolderOnDisk = SetupUI.AskOpenQuestion("Directory where to store files, empty to use in memory volatile storage", "", optional: true);
-
         AppSettings.Change(x => x.Services[ServiceName] = new Dictionary<string, object>
         {
             { "StorageType", string.IsNullOrEmpty(queueFolderOnDisk) ? "Volatile" : "Disk" },
--- a/tools/InteractiveSetup/Main.cs
+++ b/tools/InteractiveSetup/Main.cs
@@ -628,8 +628,7 @@ public static class Main
 
         AppSettings.Change(x => x.Services[ServiceName] = new Dictionary<string, object>
         {
-            { "StorageType", string.IsNullOrEmpty(queueFolderOnDisk) ? "Volatile" : "Disk" },
-            { "Directory", queueFolderOnDisk }
+            { "Directory", SetupUI.AskOpenQuestion("Directory where to store queue messages", config["Directory"].ToString()) }
         });
     }
 
--- a/tools/InteractiveSetup/Main.cs
+++ b/tools/InteractiveSetup/Main.cs
@@ -626,8 +626,6 @@ public static class Main
             };
         }
 
-        var queueFolderOnDisk = SetupUI.AskOpenQuestion("Directory where to store queue messages, empty to use in memory volatile queue", "", optional: true);
-
         AppSettings.Change(x => x.Services[ServiceName] = new Dictionary<string, object>
         {
             { "StorageType", string.IsNullOrEmpty(queueFolderOnDisk) ? "Volatile" : "Disk" },
--- a/tools/InteractiveSetup/Main.cs
+++ b/tools/InteractiveSetup/Main.cs
@@ -734,8 +734,7 @@ public static class Main
 
         AppSettings.Change(x => x.Services[ServiceName] = new Dictionary<string, object>
         {
-            { "StorageType", string.IsNullOrEmpty(queueFolderOnDisk) ? "Volatile" : "Disk" },
-            { "Directory", queueFolderOnDisk }
+            { "Directory", SetupUI.AskOpenQuestion("Directory where to store files", config["Directory"].ToString()) }
         });
     }
 
--- a/tools/InteractiveSetup/Main.cs
+++ b/tools/InteractiveSetup/Main.cs
@@ -860,8 +860,6 @@ public static class Main
             };
         }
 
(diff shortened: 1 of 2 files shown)


alkampfergit:
This will fix #358

tools/InteractiveSetup/Main.cs
```diff
@@ -622,12 +622,13 @@ private static void SimpleQueuesSetup()
             config = new Dictionary<string, object>
             {
                 { "Directory", "" },
-                { "StorageType", "Volatile" }
```
dluc, line 625: Volatile is default mode when not specified, it's on purpose. We can extend the setup wizard to ask, but we won't change the default because it would impact other things.

service/Core/Handlers/DependencyInjection.cs
```diff
@@ -103,8 +103,7 @@ public static IServiceCollection AddDefaultHandlersAsHostedServices(this IServic
         // Build generic type: HandlerAsAHostedService<THandler>
         Type handlerAsAHostedServiceTHandler = typeof(HandlerAsAHostedService<>).MakeGenericType(tHandler);
 
-        Func<IServiceProvider, IHostedService> implementationFactory =
-            serviceProvider => (IHostedService)ActivatorUtilities.CreateInstance(serviceProvider, handlerAsAHostedServiceTHandler, stepName);
+        IHostedService implementationFactory(IServiceProvider serviceProvider) => (IHostedService)ActivatorUtilities.CreateInstance(serviceProvider, handlerAsAHostedServiceTHandler, stepName);
```
dluc, line 106: what changed here?

alkampfergit: Created a private function, old code does not compile in my machine so I applied the refactoring to have it compile.

tools/InteractiveSetup/Main.cs
alkampfergit: What I do not understand is:

I started the wizard, I select to use simple storage, or simple queue, the wizard ask me the path on disk, then the wizard write that path on the configuration, but since the default is Volatile, the disk was never used. Then the user look at the configuration file and see that the file correctly contains the path, but nevertheless nothing get written. I find it confusing.

What is the purpose of asking the path to the user if then the configuration is Volatile? 

A better option would be to ask the user if they want to use in-memory (volatile) or disk.

service/Core/Handlers/DependencyInjection.cs
marcominerva: @alkampfergit see also this: https://github.com/microsoft/kernel-memory/pull/362#discussion_r1523409039

alkampfergit: Removed the modified file, after all there is no correlation with this PR it helps reduce the scope.

tools/InteractiveSetup/Main.cs
dluc: the setup wizard doesn't say path "on disk" :-)

> Directory where to store queue messages [current: _tmp/queues]:
> Directory where to store files [current: _tmp/files]:

SimpleStorage, SimpleQueue and SimpleVector internally use a filesystem that supports directories. The FS can be Persistent of Volatile, in both cases it uses "directories".  

By default it's a volatile FS, and the setup wizard doesn't ask, which is the part that could be improved.

dluc: > I've made a change: Now the user can specify an empty path to explicitly choose Volatile, but if they specify a valid path, Disk configuration is used.

This won't work, even if volatile the path cannot be empty. As I mentioned, we won't change the default, but we can improve the wizard to ask.

dluc (changes requested):
Both persistent and volatile storage use directories. Please undo changes to Main.cs, or extend the wizard to ask the storage type.

tools/InteractiveSetup/Main.cs
alkampfergit: Ok understood, if you want I can change the wizard to explicitly ask if the storage must be volatile or disk, it should be a quick and simple change.

I noticed now that the question is about path and did not tell Path on disk :D :D :), but nevertheless path lead to think that this will be saved on disk. An explicit question about

Do you want to persist on disk or you want it to be volatile? 

Should suffice

dluc: your call, if you have time to add the new question it would certainly be great :-)

tools/InteractiveSetup/Main.cs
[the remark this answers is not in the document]
dluc: I'd like to merge the fix on AppSettings.cs, so I will roll back these changes and merge. If you'd like to extend the config wizard asking for storage type we can do that on a separate PR. Thanks

tools/InteractiveSetup/Main.cs
```diff
@@ -626,9 +626,12 @@ private static void SimpleQueuesSetup()
             };
         }
 
+        var queueFolderOnDisk = SetupUI.AskOpenQuestion("Directory where to store queue messages, empty to use in memory volatile queue", "", optional: true);
+
         AppSettings.Change(x => x.Services[ServiceName] = new Dictionary<string, object>
         {
-            { "Directory", SetupUI.AskOpenQuestion("Directory where to store queue messages", config["Directory"].ToString()) }
+            { "StorageType", string.IsNullOrEmpty(queueFolderOnDisk) ? "Volatile" : "Disk" },
+            { "Directory", queueFolderOnDisk }
```
dluc, line 634:
```suggestion
            { "Directory", SetupUI.AskOpenQuestion("Directory where to store queue messages", config["Directory"].ToString()) }
```

tools/InteractiveSetup/Main.cs
```diff
@@ -626,9 +626,12 @@ private static void SimpleQueuesSetup()
             };
         }
 
+        var queueFolderOnDisk = SetupUI.AskOpenQuestion("Directory where to store queue messages, empty to use in memory volatile queue", "", optional: true);
+
```
dluc, line 630:
```suggestion
```

tools/InteractiveSetup/Main.cs
```diff
@@ -732,9 +735,12 @@ private static void SimpleFileStorageSetup()
             };
         }
 
+        var queueFolderOnDisk = SetupUI.AskOpenQuestion("Directory where to store files, empty to use in memory volatile storage", "", optional: true);
+
```
dluc, line 739:
```suggestion
```

tools/InteractiveSetup/Main.cs
```diff
@@ -860,9 +866,12 @@ private static void SimpleVectorDbSetup(bool force = false)
             };
         }
 
+        var queueFolderOnDisk = SetupUI.AskOpenQuestion("Directory where to store vectors, empty to use in memory volatile memory", "", optional: true);
+
```
dluc, line 870:
```suggestion
```

tools/InteractiveSetup/Main.cs
```diff
@@ -732,9 +735,12 @@ private static void SimpleFileStorageSetup()
             };
         }
 
+        var queueFolderOnDisk = SetupUI.AskOpenQuestion("Directory where to store files, empty to use in memory volatile storage", "", optional: true);
+
         AppSettings.Change(x => x.Services[ServiceName] = new Dictionary<string, object>
         {
-            { "Directory", SetupUI.AskOpenQuestion("Directory where to store files", config["Directory"].ToString()) }
+            { "StorageType", string.IsNullOrEmpty(queueFolderOnDisk) ? "Volatile" : "Disk" },
+            { "Directory", queueFolderOnDisk }
```
dluc, line 743:
```suggestion
            { "Directory", SetupUI.AskOpenQuestion("Directory where to store files", config["Directory"].ToString()) }
```

tools/InteractiveSetup/Main.cs
```diff
@@ -860,9 +866,12 @@ private static void SimpleVectorDbSetup(bool force = false)
             };
         }
 
+        var queueFolderOnDisk = SetupUI.AskOpenQuestion("Directory where to store vectors, empty to use in memory volatile memory", "", optional: true);
+
         AppSettings.Change(x => x.Services[ServiceName] = new Dictionary<string, object>
         {
-            { "Directory", SetupUI.AskOpenQuestion("Directory where to store vectors", config["Directory"].ToString()) }
+            { "StorageType", string.IsNullOrEmpty(queueFolderOnDisk) ? "Volatile" : "Disk" },
+            { "Directory", queueFolderOnDisk }
```
dluc, line 874:
```suggestion
            { "Directory", SetupUI.AskOpenQuestion("Directory where to store vectors", config["Directory"].ToString()) }
```

dluc (changes requested):

Issue #361, opened 2024-03-13
[Bug] NotSupportedException when referencing Microsoft.SemanticKernel library v1.6.1


marcominerva:
### Context / Scenario

Kernel Memory (in particular the **AzureOpenAI** project) has a dependency on **Microsoft.SemanticKernel.Connectors.OpenAI v1.4**.

If we add the **KernelMemory.Core** NuGet Package to a project, and then explicitly add a reference to **Microsoft.SemanticKernel v1.6.1**, the `KernelMemoryBuilder.Build` methods fails with a `NotSupportedException `exception

### What happened?

Digging in the source code, the problem is caused, for example, in the `AzureOpenAITextGenerator` class, in particular when it instantiate the `OpenAIClientOptions` object:

https://github.com/microsoft/kernel-memory/blob/8aa175443d24d212d42189a16e1bc832e6800a40/extensions/AzureOpenAI/AzureOpenAITextGenerator.cs#L69-L77

The constructor of this class, that is contained in the **Azure.AI.OpenAI** library, uses an Enum to determine with API version to use. The verison used by **KernelMemory.AzureOpenAI** (beta13) causes a mismatch with the version used by **Microsoft.SemanticKernel v1.6.1** (beta14), resulting in a `NotSupportedException`.

### Importance

I cannot use Kernel Memory

### Platform, Language, Versions

KernelMemory v 0.33.240312.1
SemanticKernel v1.6.1

Ivaznzhou:
Reflector Microsoft.KernelMemory.AI.OpenAI, Version=0.33.0.0, Culture=neutral, PublicKeyToken=f300afd708cefcd3
Code 
OpenAIClientOptions options1 = new OpenAIClientOptions(OpenAIClientOptions.ServiceVersion.V2024_02_15_Preview | OpenAIClientOptions.ServiceVersion.V2023_05_15);

OpenAIClientOptions.ServiceVersion.V2024_02_15_Preview | OpenAIClientOptions.ServiceVersion.V2023_05_15 =7
But 
Azure.AI.OpenAI OpenAIClientOptions.ServiceVersion no 7

Ivaznzhou:
screenshot
![image](https://github.com/microsoft/kernel-memory/assets/83056816/16f31618-b32a-4d44-b4e3-0c5d3bd02a22)

dluc:
[...]

PR #362, opened 2024-03-13
Bump SemanticKernel dependencies to v1.6.1

marcominerva:
This PR solves a version mismatch between libraries. See https://github.com/microsoft/kernel-memory/issues/361.

This PR updates the package dependencies to correctly support the latest version of **Azure.AI.OpenAI library**, that is required to use the right API version of the service.

### DIFF
--- a/service/Core/Handlers/DependencyInjection.cs
+++ b/service/Core/Handlers/DependencyInjection.cs
@@ -103,11 +103,11 @@ public static partial class DependencyInjection
         // Build generic type: HandlerAsAHostedService<THandler>
         Type handlerAsAHostedServiceTHandler = typeof(HandlerAsAHostedService<>).MakeGenericType(tHandler);
 
-        IHostedService implementationFactory(IServiceProvider serviceProvider)
+        IHostedService ImplementationFactory(IServiceProvider serviceProvider)
             => (IHostedService)ActivatorUtilities.CreateInstance(serviceProvider, handlerAsAHostedServiceTHandler, stepName);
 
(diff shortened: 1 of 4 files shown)


service/Core/Handlers/DependencyInjection.cs
```diff
@@ -103,8 +103,8 @@ public static IServiceCollection AddDefaultHandlersAsHostedServices(this IServic
         // Build generic type: HandlerAsAHostedService<THandler>
         Type handlerAsAHostedServiceTHandler = typeof(HandlerAsAHostedService<>).MakeGenericType(tHandler);
 
-        Func<IServiceProvider, IHostedService> implementationFactory =
-            serviceProvider => (IHostedService)ActivatorUtilities.CreateInstance(serviceProvider, handlerAsAHostedServiceTHandler, stepName);
+        IHostedService implementationFactory(IServiceProvider serviceProvider)
```
dluc, line 106: why this change?

marcominerva: Because trying to rebuild the solution (using **KernelMemory.sln**, `main` branch), I get this compilation error:

image

And the message point me to this page: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0039. I just updated Visual Studio 2022 to v17.9.3, perhaps it is a new analyzer?

dluc: looks like a new analyzer only in VS, odd

dluc (approved):

PR #370, opened 2024-03-18
Fix broken link

pascalberger:
Link to security filters is broken

Fixes link to security filters when deployed to https://microsoft.github.io/


dluc (approved):
thank you!

PR #377, opened 2024-03-21
Fix README link

Valkozaur:
The link was corrupted and not aligning with the current examples code structure.
I applied a fix.


dluc (approved):
thanks!

PR #384, opened 2024-03-25
Updated MSExcelDecoder to handle blank cell values

neel015:
Objective is to support excel files which has few blank values within the used range.
Taking example of below excel
image

- Currently the code ignores the blank values, and that creates distortion in the arrangement of values.
Output of the above excel file looks something like below
```
"EmployeeName", "EmployeeSalary", "EmployeeDepartment", "EmployeeCity", "EmployeeId"
"John", 1000, "Accounts", "New York", 1
"Jack", 2000, "Tax", "Bangalore", 2
"Ryan", 3000, "HR", "Tokyo", 3
"Rob", 2000, "Tax", 4
"Pablo", 5000, "Accounts", "Paris", 
```
(See the employee city for Rob has been substituted by the employee Id 4)

- With the proposed change, a default value `Blank` will be substituted when these kinds of scenarios are observed.  Consumers are free to overwrite the default value with the help of the newly introduced constructor variable

```

"EmployeeName", "EmployeeSalary", "EmployeeDepartment", "EmployeeCity", "EmployeeId"
"John", 1000, "Accounts", "New York", 1
"Jack", 2000, "Tax", "Bangalore", 2
"Ryan", 3000, "HR", "Tokyo", 3
"Rob", 2000, "Tax", Blank, 4
"Pablo", 5000, "Accounts", "Paris", 5
```
(See the employee city for Rob is substituted with `Blank`)

### DIFF
--- a/service/Core/DataFormats/Office/MsExcelDecoder.cs
+++ b/service/Core/DataFormats/Office/MsExcelDecoder.cs
@@ -26,6 +26,7 @@ public class MsExcelDecoder
     private readonly string _columnSeparator;
     private readonly string _rowSuffix;
     private readonly string _blankCellValue;
+
     public MsExcelDecoder(
         bool withWorksheetNumber = true,
         bool withEndOfWorksheetMarker = false,
--- a/service/Core/DataFormats/Office/MsExcelDecoder.cs
+++ b/service/Core/DataFormats/Office/MsExcelDecoder.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft. All rights reserved.
+﻿// Copyright (c) Microsoft. All rights reserved.
 
 using System;
 using System.IO;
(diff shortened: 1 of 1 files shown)


service/Core/DataFormats/Office/MsExcelDecoder.cs
```diff
@@ -15,6 +15,7 @@ public class MsExcelDecoder
     private const string DefaultRowPrefix = "";
     private const string DefaultColumnSeparator = ", ";
     private const string DefaultRowSuffix = "";
+    private const string DefaultBlankCellValue = "Blank";
```
majern, line 18: Why not to have DefaultBlankCellValue just `string.Empty`?

service/Core/DataFormats/Office/MsExcelDecoder.cs
```diff
@@ -24,7 +25,7 @@ public class MsExcelDecoder
     private readonly string _rowPrefix;
     private readonly string _columnSeparator;
     private readonly string _rowSuffix;
-
+    private readonly string _blankCellValue;
     public MsExcelDecoder(
```
dluc, line 29:
```suggestion

    public MsExcelDecoder(
```

dluc (approved):

Issue #289, opened 2024-02-02
ImportDocumentAsync fails to process large documents with Azure.RequestFailedException: You are sending too many requests. Please try again later.


johnnyreilly:
### Context / Scenario

First of all, thanks for KernelMemory - it's brilliant!  We've been building with it and it's been a game changer in terms of the things it makes possible.

In our scenario, we have users that want to chat to very large documents; like 600 page PDFs. We attempt to process such a document like so (C#):

```cs
await _memory.ImportDocumentAsync(content: documentContent, fileName: fileName, documentId: documentId, index: index, tags: tags);
```

### What happened?

The `ImportDocumentAsync` runs for around 10 minutes and then throws an `Azure.RequestFailedException` with the message "You are sending too many requests. Please try again later.".  See stack trace below:

```
2024-02-01T20:06:55.767571991Z       Error processing document: DocumentToProcess { DocumentUrl = https://stourgptprod.blob.core.windows.net/ourgpt/Big%20Book.pdf, AssistantName = ourgpt, AssistantDisplayName = OurGPT, Description = Group Security assistant, MaintainerEmails = System.String[], AdGroupIds = System.String[], User = UserEmailAndId  }
2024-02-01T20:06:55.767584749Z       System.Exception: Problem chunking document: https://stourgptprod.blob.core.windows.net/ourgpt/Big%20Book.pdf
2024-02-01T20:06:55.767601865Z        ---> Azure.RequestFailedException: You are sending too many requests. Please try again later.
2024-02-01T20:06:55.767607622Z       Status: 503 (Service Unavailable)
2024-02-01T20:06:55.767611742Z       
2024-02-01T20:06:55.767616947Z       Content:
2024-02-01T20:06:55.767621400Z       {"error":{"code":"","message":"You are sending too many requests. Please try again later."}}
2024-02-01T20:06:55.767625940Z       
2024-02-01T20:06:55.767630851Z       Headers:
2024-02-01T20:06:55.767637232Z       Server: Microsoft-IIS/10.0
2024-02-01T20:06:55.767641552Z       Strict-Transport-Security: REDACTED
2024-02-01T20:06:55.767645998Z       Preference-Applied: REDACTED
2024-02-01T20:06:55.767650369Z       throttle-reason: rateLimitExceeded
2024-02-01T20:06:55.767654576Z       client-request-id: f668f002-39c1-42ef-aca3-e47b7feb4cc6
[...]

PR #387, opened 2024-03-27
Check index once per unique index name during save embeddings.

spenavajr:

I had the same problem as the submitter of this issue: https://github.com/microsoft/kernel-memory/issues/289. For every embedding during the save process, a call is made to list indexes. Using Azure AI Search, I was getting rate limit errors, i.e. `Azure.RequestFailedException with the message "You are sending too many requests. Please try again later."`

Made an adjustment to simply cache indexes that were already checked/created during the save embeddings process. This skips the unnecessary call to ListIndexes for every single embedding.

### DIFF
--- a/service/Core/Handlers/SaveRecordsHandler.cs
+++ b/service/Core/Handlers/SaveRecordsHandler.cs
@@ -143,7 +143,9 @@ public class SaveRecordsHandler : IPipelineStepHandler
 
             foreach (IMemoryDb client in this._memoryDbs)
             {
-                if (!createdIndexes.Contains(pipeline.Index))
+                var key = GetPipelineKeyForSave(client, pipeline.Index);
+
+                if (!createdIndexes.Contains(key))
                 {
                     this._log.LogTrace("Creating index '{0}'", pipeline.Index);
                     await client.CreateIndexAsync(pipeline.Index, record.Vector.Length, cancellationToken).ConfigureAwait(false);
@@ -212,7 +214,9 @@ public class SaveRecordsHandler : IPipelineStepHandler
 
                     foreach (IMemoryDb client in this._memoryDbs)
                     {
-                        if (!createdIndexes.Contains(pipeline.Index))
+                        var key = GetPipelineKeyForSave(client, pipeline.Index);
+
+                        if (!createdIndexes.Contains(key))
                         {
                             this._log.LogTrace("Creating index '{0}'", pipeline.Index);
                             await client.CreateIndexAsync(pipeline.Index, record.Vector.Length, cancellationToken).ConfigureAwait(false);
@@ -281,6 +285,11 @@ public class SaveRecordsHandler : IPipelineStepHandler
         return pipeline.Files.SelectMany(f1 => f1.GeneratedFiles.Where(
                 f2 => f2.Value.ArtifactType == DataPipeline.ArtifactTypes.TextPartition || f2.Value.ArtifactType == DataPipeline.ArtifactTypes.SyntheticData)
             .Select(x => new FileDetailsWithRecordId(pipeline, x.Value)));
+    }
+
+    private static string GetPipelineKeyForSave(IMemoryDb client, string index)
+    {
+        return $"{client.GetType().Name}::{index}";
     }
 
     private async Task<string> GetSourceUrlAsync(
--- a/service/Core/Handlers/SaveRecordsHandler.cs
+++ b/service/Core/Handlers/SaveRecordsHandler.cs
@@ -149,7 +149,7 @@ public class SaveRecordsHandler : IPipelineStepHandler
                 {
                     this._log.LogTrace("Creating index '{0}'", pipeline.Index);
                     await client.CreateIndexAsync(pipeline.Index, record.Vector.Length, cancellationToken).ConfigureAwait(false);
-                    createdIndexes.Add(pipeline.Index);
+                    createdIndexes.Add(key);
                 }
 
                 this._log.LogTrace("Saving record {0} in index '{1}'", record.Id, pipeline.Index);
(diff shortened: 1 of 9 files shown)


luismanez:
Thanks @spenavajr 

There is also another call to ```client.CreateIndexAsync(...)``` in the method:

```csharp
public async Task<(bool success, DataPipeline updatedPipeline)> SavePartitionsAsync(
```

I'm pretty sure we also need your fix in that method. Aas far as I understand, there is one method that only Insert the embeddings data, and another one the rest of the fields (tags, payload).

Could you add your fix to that method, please?

spenavajr:

spenavajr:
@luismanez you are correct, I overlooked that. Thank you!

luismanez:
> @luismanez you are correct, I overlooked that. Thank you! Updated.

Thanks again. However, looking at it again, I think we still need another change. 

The exiting code iterates all the "chunks", and foreach item, iterates all the configured Memories (apparently you can configure multiple Memories, like one for Az Search, another one for Postgree, etc). If we only have one Memory, then your code will work, but if we have multiple memories, as the ```createdIndexes``` is only storing the name of the Index, when it enters in the 2nd memory item, the Index won't be created for that 2nd memory, and when it calls ```client.UpsertAsync``` will fail.

I don't see an easy solution for this, unless you move the creation before the initial foreach, but you should then to check multiple things...

@dluc Could you explain why we have multiple Memories? I don't see how to configure more than one. Initially, the ```KernelMemoryBuilder``` adds the ```SimpleVectorDb``` memory, and then, all the Extensions that I can see, like ```AddAzureAISearchAsMemoryDb```, ```AddRedisAsMemoryDb```, etc, register the Memory as Singleton:
```.AddSingleton<IMemoryDb, RedisMemory>();``` The only way seems is creating the BaseOrchestrator manually...

**Comment by spenavajr** [comment]
@luismanez that's a good point, thanks for digging in. I would assume having multiple memory databases is to simply store those memories in multiple places?

One suggestion is - we could modify what goes into the `createdIndexes` hash set to include a memory client prefix, e.g.

`AzureAISearch::<index name>` surely that should solve the problem, no?

**Comment by luismanez** [comment]
> @luismanez that's a good point, thanks for digging in. I would assume having multiple memory databases is to simply store those memories in multiple places?
> 
> One suggestion is - we could modify what goes into the `createdIndexes` hash set to include a memory client prefix, e.g.
> 
> `AzureAISearch::<index name>` surely that should solve the problem, no?

yup, but you just have an ```IMemoryDb``` object, and I haven't found any kind of id in that interface that we can use for composing something like a primaryKey...

**Comment by luismanez** [comment]
@spenavajr might work with calling the GetType() method??

**Comment by spenavajr** [comment]
> @spenavajr might work with calling the GetType() method??

Yeah, I was thinking `client.GetType().Name`

I will try it out & report back.

**Comment by spenavajr** [comment]
@luismanez seems to be working, check latest update - the team may want this designed differently... but it seems to solve the original issue at hand for now (I may fork this and use it for a while if not merged soon 😄)

**Comment by luismanez** [comment]
Looks good!! Thanks a lot.

@dluc please, we need this fix otherwise we cannot work with long files when using Az Search as Memory. Thanks!

**Comment by dluc** [comment]
> Looks good!! Thanks a lot.
> 
> @dluc please, we need this fix otherwise we cannot work with long files when using Az Search as Memory. Thanks!

someone will look into this asap, however, please note that due to some constraints, the earliest we might be able to address this could be after April 8th

File: service/Core/Handlers/SaveRecordsHandler.cs
Comment (dluc) [inline service/Core/Handlers/SaveRecordsHandler.cs:1]: This might work, though I would handle the exceptions resulting from "index not found" and ensure the index is created (logging warnings etc)

File: service/Core/Handlers/SaveRecordsHandler.cs [hunk 1]
```diff
@@ -273,6 +287,11 @@ private static IEnumerable<FileDetailsWithRecordId> GetListOfPartitionAndSynthet
             .Select(x => new FileDetailsWithRecordId(pipeline, x.Value)));
     }
 
+    private static string GetPipelineKeyForSave(IMemoryDb client, string index)
```
dluc, line 290: The function name is a bit odd... 

It's not clear why the code uses the client class name, could you add a comment? I would consider using `ConditionalWeakTable` so the code will scale to multiple instances of the same type (connected to different endpoints)

dluc:
> @dluc Could you explain why we have multiple Memories? I don't see how to configure more than one. Initially, the `KernelMemoryBuilder` adds the `SimpleVectorDb` memory, and then, all the Extensions that I can see, like `AddAzureAISearchAsMemoryDb`, `AddRedisAsMemoryDb`, etc, register the Memory as Singleton:
> `.AddSingleton<IMemoryDb, RedisMemory>();` The only way seems is creating the BaseOrchestrator manually...

The intention is to allow combining different storage types that offer different cost, performance, features:

```
      // Vectors can be written to multiple storages, e.g. for data migration, A/B testing, etc.
      "MemoryDbTypes": [
        "SimpleVectorDb"
      ],
```

dluc (changes requested):
Will revisit after the comments have been addressed. Thanks!

spenavajr:
@dluc been super busy at work, will get on this as soon as I can!

service/Core/Handlers/SaveRecordsHandler.cs
dluc: I added part of the IndexNotFound logic in the handler and in AzureSearchMemory. The same should be implement in other memory connectors

service/Core/Handlers/SaveRecordsHandler.cs
dluc: added TODO for future wok

dluc (approved):

Issue #379, opened 2024-03-21
[Feature Request] Configure Decoders with Dependency Injection


marcominerva:
### Context / Scenario

This is related to https://github.com/microsoft/kernel-memory/issues/375. If we want to customize how content is extracted, we can create a new implementation of the [TextExtractionHandler](https://github.com/microsoft/kernel-memory/blob/main/service/Core/Handlers/TextExtractionHandler.cs). However, in most cases, the default handler is fine, and we just need to customize how the text is actually retrieved (i.e., applying some custom processing).

### The problem

Currently, **Decoders** are explicitly created in the [TextExtractionHandler.InvokeAsync](https://github.com/microsoft/kernel-memory/blob/main/service/Core/Handlers/TextExtractionHandler.cs#L55) method. So, if we want to customize the extraction step for a particular file type (or even add support for a new file type), the only way is to provide a whole implementation of the `TextExtractionHandler` itself.

### Proposed solution

If **Decoders** are registered in the Service Collection, it will be easy to define a custom implementation for a particular MIME type (for example, if we want to customize how text is extracted from a PDF file), without the need to redefine the whole handler.

Of course, this won't introduce any breaking change, because current **Decoders** will be registered as default, but we can easily create our own if needed.

Moreover, with this approach, we can dynamically add support for new file types just adding a new registration in the Service Collection, without touching the `TextExtractionHandler` at all.

@dluc, If you think that this feature is interesting, I will be more than happy to take care of the PR 😀

### Importance

would be great to have

dluc:
I see the benefits, before jumping to it, could you describe what approach you would take, e.g. in terms of classes and interfaces?

marcominerva:
Sure!

I was thinking about adding an interface in the `Abstractions` project like the following (with `async` methods to handle all the scenarios):

```csharp
public interface IDecoder
{
    IEnumerable<string> SupportedMimeTypes { get; }

    Task<FileContent> ExtractContentAsync(string filename);

    Task<FileContent> ExtractContentAsync(BinaryData data);

    Task<FileContent> ExtractContentAsync(Stream data);
}
```

Then, an extension method for the `KernelMemoryBuilder`:

```csharp
public static IKernelMemoryBuilder WithDecoder<T>(this IKernelMemoryBuilder builder) where T : class, IDecoder
{
    builder.Services.AddSingleton<IDecoder, T>();
    return builder;
}
```

[...]

PR #398, opened 2024-04-10
Support for custom Content Decoders

marcominerva:
See https://github.com/microsoft/kernel-memory/issues/379.

Compared to what described in https://github.com/microsoft/kernel-memory/issues/379, I had to slightly modify the `IContentDecoder` interface:

```csharp
public interface IContentDecoder
{
    bool SupportsMimeType(string mimeType);

    Task<FileContent?> ExtractContentAsync(string handlerStepName, 
        DataPipeline.FileDetails file, string filename, CancellationToken cancellationToken = default);

    Task<FileContent?> ExtractContentAsync(string handlerStepName,
        DataPipeline.FileDetails file, BinaryData data, CancellationToken cancellationToken = default);

    Task<FileContent?> ExtractContentAsync(string handlerStepName, 
        DataPipeline.FileDetails file, Stream data, CancellationToken cancellationToken = default);
}
```

I do this because come decoders need to access information and to log to the `DataPipeline` (see for example the Word decoder: https://github.com/marcominerva/kernel-memory/blob/9dd1c17c075bc209dca4fccb28cdba615e4185e8/service/Core/DataFormats/Office/MsWordDecoder.cs#L42-L51).


service/Abstractions/Pipeline/DataPipeline.cs
```diff
@@ -149,12 +149,19 @@ public void MarkProcessedBy(IPipelineStepHandler handler)
...
+        /// <summary>
+        /// Add a new log entry, with some important information for the end user.
+        /// DO NOT STORE PII OR SECRETS here.
+        /// </summary>
+        /// <param name="handlerStepName">Step name of the handler sending the information to log</param>
+        /// <param name="text">Text to store for the end user</param>
+        public void Log(string handlerStepName, string text)
+        {
+            this.LogEntries ??= new List<PipelineLogEntry>();
+            this.LogEntries.Add(new PipelineLogEntry(source: handlerStepName, text: text));
```
dluc, line 164: I would remove this, e.g. decoders should not know anything about the pipeline and it's internal logging. If needed, we can include log data in FileContent, or return log data as a separate object.

service/Core/DataFormats/Office/MsExcelDecoder.cs
```diff
@@ -1,70 +1,59 @@
...
-        return this.ExtractContent(stream);
+        return this.ExtractContentAsync(handlerStepName, file, stream, cancellationToken);
     }
 
-    public FileContent ExtractContent(Stream data)
+    public Task<FileContent?> ExtractContentAsync(string handlerStepName, DataPipeline.FileDetails file, Stream data, CancellationToken cancellationToken = default)
     {
+        if (file.MimeType == MimeTypes.MsExcel)
+        {
+            file.Log(
```
dluc, line 46: I would remove this, e.g. decoders should not know anything about the pipeline, file details, and internal logging logic. If needed, we can include log data in FileContent, or return some log data as a separate object.

Directory.Packages.props
```diff
@@ -8,9 +8,9 @@
     <PackageVersion Include="Azure.Search.Documents" Version="11.5.1" />
     <PackageVersion Include="Azure.Storage.Blobs" Version="12.19.1" />
     <PackageVersion Include="Azure.Storage.Queues" Version="12.17.1" />
-    <PackageVersion Include="ClosedXML" Version="0.102.1" />
+    <PackageVersion Include="ClosedXML" Version="0.102.2" />
```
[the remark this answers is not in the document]
marcominerva: According to the release notes at https://github.com/ClosedXML/ClosedXML/releases/tag/0.102.2, the only change in this version is a warning for the supported version of DocumentFormat.OpenXml library.

service/Abstractions/DataFormats/IContentDecoder.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Threading.Tasks;
+using Microsoft.KernelMemory.Pipeline;
+
+namespace Microsoft.KernelMemory.DataFormats;
+
+public interface IContentDecoder
+{
+    IEnumerable<string> SupportedMimeTypes { get; }
+
+    Task<FileContent?> ExtractContentAsync(string handlerStepName, DataPipeline.FileDetails file, string filename, CancellationToken cancellationToken = default);
```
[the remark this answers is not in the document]
marcominerva: It is necessary so that the **TextExtractionHandler** knows that the file has been skipped.

service/Core/Handlers/TextExtractionHandler.cs
```diff
@@ -131,142 +127,61 @@ public class TextExtractionHandler : IPipelineStepHandler
         CancellationToken cancellationToken)
     {
         bool skipFile = false;
-        var content = new FileContent();
-        string extractType = MimeTypes.PlainText;
-
-        switch (uploadedFile.MimeType)
+        FileContent? content = null;
+        string extractType = uploadedFile.MimeType switch
```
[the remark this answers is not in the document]
marcominerva: Yes, we could add a MimeType property that values `PlainText` as default, and change it when necessary. What do you think about?

service/Core/DataFormats/Office/MsExcelDecoder.cs
marcominerva: I think that adding adding log data in `FileContent` could be a good solution. As we need this for logging the reason of  skipping the file, it could be a simple `SkippingReason` string property. Or do you have in mind something more structured?

service/Abstractions/Pipeline/DataPipeline.cs
marcominerva: I think that adding adding log data in `FileContent` could be a good solution. As we need this for logging the reason of  skipping the file, it could be a simple `SkippingReason` string property. Or do you have in mind something more structured?

service/Core/DataFormats/Office/MsExcelDecoder.cs
dluc: I would leave it open, a list of entries, reusing a struct/class similar to PipelineLogEntry, without using the same class though, at the cost of duplicating it. Something like:

```csharp
public class FileContent
{
    public sealed class LogEntry
    {
        [JsonPropertyOrder(0)]
        [JsonPropertyName("t")]
        public DateTimeOffset Time { get; set; } = DateTimeOffset.UtcNow;

        [JsonPropertyOrder(1)]
        [JsonPropertyName("src")]
        public string Source { get; set; }

        [JsonPropertyOrder(2)]
        [JsonPropertyName("txt")]
        public string Text { get; set; }

        public LogEntry(string source, string text)
        {
            this.Source = source;
            this.Text = text;
        }
    }
    
    [JsonPropertyOrder(0)]
    [JsonPropertyName("sections")]
    public List<FileSection> Sections { get; set; } = new();

    [JsonPropertyOrder(100)]
    [JsonPropertyName("log")]
    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
    public List<LogEntry>? LogEntries { get; set; } = null;
}
```

Directory.Packages.props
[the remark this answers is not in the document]
dluc: unless it's strictly necessary I'd leave this out, so we won't have to test for regressions

service/Abstractions/DataFormats/IContentDecoder.cs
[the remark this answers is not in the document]
dluc: what about using a specific exception instead? a nullable object allows to define only a single scenario (skipped in this case), but with exceptions we would allow defining also reasons for skipping and other scenarios, e.g. file too big, timeouts, etc.

service/Core/DataFormats/Office/MsExcelDecoder.cs
dluc: Alternatively, if there's just one log entry to add, I would just use the exception approach mentioned in the other comment, including in the exception all the details to be logged by the handler. That might be just easier and less code

service/Abstractions/DataFormats/IContentDecoder.cs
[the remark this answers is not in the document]
marcominerva: Initially I was thinking about exceptions, but I have tried to avoid them for performance reasons. But we could definitely use them. In fact, using exceptions we could also solve the logging problems described in the later comments.

service/Core/DataFormats/Office/MsExcelDecoder.cs
marcominerva: Yes, I think that using exception is the most common way to handle this situation.

service/Abstractions/Abstractions.csproj
```diff
@@ -2,30 +2,31 @@
 
     <PropertyGroup>
         <TargetFramework>netstandard2.0</TargetFramework>
+        <LangVersion>latest</LangVersion>
```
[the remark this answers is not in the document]
dluc: Abstractions is designed to be .NET Standard2.0 compatible and there are some breaking changes that I haven't reviewed in v12. 

The SupportedMimeType initialization is used only in Core btw. If you're looking for a concise syntax, changing

```csharp
IEnumerable<string> SupportedMimeTypes
```
to
```csharp
List<string> SupportedMimeTypes
```

you can replace
```csharp
IEnumerable<string> SupportedMimeTypes { get; } = [MimeTypes.MsExcelX, MimeTypes.MsExcel];
```
with
```csharp
List<string> SupportedMimeTypes { get; } = new() { MimeTypes.MsExcelX, MimeTypes.MsExcel };
```

KernelMemory.sln
dluc, line 1: a lot of changes here, do you know which ones should be merged?

examples/205-dotnet-extract-text-from-docs/Program.cs
```diff
@@ -3,16 +3,20 @@
 using Microsoft.KernelMemory.DataFormats;
 using Microsoft.KernelMemory.DataFormats.Office;
 using Microsoft.KernelMemory.DataFormats.Pdf;
+using Microsoft.KernelMemory.Pipeline;
 
-FileContent content = new();
+var content = new FileContent();
```
dluc, line 8:
```suggestion
FileContent content = new();
```

KernelMemory.sln
marcominerva: I just added the example **108-dotnet-custom-content-decoders**, I don't know why Visual Studio changed all the file content.

service/Core/DataFormats/Office/MsExcelDecoder.cs
```diff
@@ -1,70 +1,54 @@
...
-    {
-        this._withWorksheetNumber = withWorksheetNumber;
-        this._withEndOfWorksheetMarker = withEndOfWorksheetMarker;
-        this._withQuotes = withQuotes;
+    private readonly MsExcelConfig _config;
+    private readonly ILogger<MsExcelDecoder> _log;
 
-        this._worksheetNumberTemplate = worksheetNumberTemplate ?? DefaultSheetNumberTemplate;
-        this._endOfWorksheetMarkerTemplate = endOfWorksheetMarkerTemplate ?? DefaultEndOfSheetTemplate;
+    public IEnumerable<string> SupportedMimeTypes { get; } = new[] { MimeTypes.MsExcelX, MimeTypes.MsExcel };
```
dluc, line 22: why add `MimeTypes.MsExcel` here if the format is not supported and the result is an exception?

service/Core/DataFormats/Office/MsPowerPointDecoder.cs
```diff
@@ -1,80 +1,55 @@
...
-
-    /// <param name="slideNumberTemplate">Template used for the optional slide number added at the start of each slide</param>
-    /// <param name="endOfSlideMarkerTemplate">Template used for the optional text added at the end of each slide</param>
-    public MsPowerPointDecoder(
-        string slideNumberTemplate = "# Slide {number}",
-        string endOfSlideMarkerTemplate = "# End of slide {number}")
+    private readonly MsPowerPointConfig _config;
+    private readonly ILogger<MsPowerPointDecoder> _log;
+
+    public IEnumerable<string> SupportedMimeTypes { get; } = new[] { MimeTypes.MsPowerPointX, MimeTypes.MsPowerPoint };
```
dluc, line 23: why add MimeTypes.MsPowerPoint here if the format is not supported and the result is an exception?

service/Core/DataFormats/Office/MsWordDecoder.cs
```diff
@@ -4,27 +4,49 @@
...
 
 namespace Microsoft.KernelMemory.DataFormats.Office;
 
-public class MsWordDecoder
+public class MsWordDecoder : IContentDecoder
 {
-    public FileContent ExtractContent(string filename)
+    private readonly ILogger<MsWordDecoder> _log;
+
+    public IEnumerable<string> SupportedMimeTypes { get; } = new[] { MimeTypes.MsWordX, MimeTypes.MsWord };
```
dluc, line 21: ditto

service/Core/DataFormats/Office/MsExcelDecoder.cs
marcominerva: Because I have kept the original behavior that uses a specific log message for old Office 97-2003 file formats: https://github.com/microsoft/kernel-memory/blob/489e4c66f96b0f3f4320d140a46d69fa65a1a26a/service/Core/Handlers/TextExtractionHandler.cs#L173. If it is OK to use the standard message, then we can safely remove this.

service/Core/DataFormats/Office/MsPowerPointDecoder.cs
marcominerva: Because I have kept the original behavior that uses a specific log message for old Office 97-2003 file formats: https://github.com/microsoft/kernel-memory/blob/489e4c66f96b0f3f4320d140a46d69fa65a1a26a/service/Core/Handlers/TextExtractionHandler.cs#L173. If it is OK to use the standard message, then we can safely remove this.

service/Core/DataFormats/Office/MsExcelDecoder.cs
dluc: yes I think log message added by the handler is sufficient when it doesn't find a handler, in which case we don't even need the exception I guess

service/Core/DataFormats/Office/MsWordDecoder.cs
marcominerva: Same as above, if you're agree I can remove all the checks for the old Office file formats and rely on the default skip message.

service/Core/DataFormats/WebPages/HtmlDecoder.cs
```diff
@@ -1,32 +1,55 @@
...
-    public FileContent ExtractContent(Stream data)
+    public Task<FileContent> ExtractContentAsync(string name, Stream data, string mimeType, CancellationToken cancellationToken = default)
     {
+        this._log.LogDebug("Extracting text from HTML file {0}", name);
+
         var doc = new HtmlDocument();
         doc.Load(data);
 
         var result = new FileContent();
+        if (mimeType == MimeTypes.MarkDown)
```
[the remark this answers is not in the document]
marcominerva: Yes, if you can take care of this change I really appreciate it. I have understand what you mean, but I'm not sure how to implement it in an efficient way ;)

KernelMemory.sln
marcominerva: Probably there are just the following lines: https://github.com/marcominerva/kernel-memory/blob/2151d4a816ee44de666231d2ec8951bd3c960b2f/KernelMemory.sln#L239-L240

service/Core/DataFormats/WebPages/HtmlDecoder.cs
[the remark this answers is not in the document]
dluc: cool, I'll wait for your changes to be complete, could you drop me a msg here to let me know, so we don't step on each other :-)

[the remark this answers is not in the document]
marcominerva: Sorry, there was an error in the code. The check for the `MarkDown` Mime type must be done in the [TextDecoder.cs class](https://github.com/marcominerva/kernel-memory/blob/0447f6b788dc6a9f10f08f86e355dd43ccec9dcc/service/Core/DataFormats/Text/TextDecoder.cs#L46-L49), because it support multiple types (`PlainText`, `MarkDown`, and `Json`). I have updated it and removed the corresponding lines from **HtmlDecoder.cs**, that supports only `Html` files.

So, `mimeType` is still a required parameter for Decoders. I think it is a good idea to keep this parameter in any case, so that, if necessary, a decoder that supports multiple Mime types can easily determine the actual type on which it is invoked.

If, however, you think it's better to remove it, I can create a specific `MarkDownDecoder` that handles only this Mime type and then set the `MimeType` property of `FileContent` to the correct value.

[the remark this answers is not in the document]
dluc: I see pros and cons:
* if we don't pass it, it will invite developers writing small decoders for a specific type, e.g. making it impossible to write a gigantic decoder that handle all types -- this would be good
* if we pass it, we should validate that it is in the list of supported types

having to make a choice, I would not pass it, and have specific decoders, easier to review and manage long term I think. In fact there's some work to do with JSON files, and having a specific decoder would help

KernelMemory.sln
dluc: fixed

service/Core/DataFormats/WebPages/HtmlDecoder.cs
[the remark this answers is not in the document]
dluc: fixed

dluc:
I applied the changes discussed earlier, though it will need some testing to make sure all is working fine in the pipeline.

Given the changes to Abstractions we'll need to split the PR in two parts, one PR for the changes to Abstractions, which needs to be released before Core can use them. The alternative would be using a bunch of `#ifdef` while transitioning, but that could make manual tests more expensive...

marcominerva:
Thank you for your support @dluc!

So, now is there something else I can do?

marcominerva:
I'm also wondering if, using this new extensible solution, the [MimeTypeDetector ](https://github.com/microsoft/kernel-memory/blob/main/service/Abstractions/Pipeline/MimeTypes.cs) should be updated as well.

Because with this PR I could for example add support for CSV file, simply adding a decoder, but then the code will throw an exception because the `text/csv` type isn't included in the list, so we need to provide also a custom implementation of **IMimeTypeDetection**.

dluc:
> I'm also wondering if, using this new extensible solution, the [MimeTypeDetector ](https://github.com/microsoft/kernel-memory/blob/main/service/Abstractions/Pipeline/MimeTypes.cs) should be updated as well.
> 
> With this PR, I could for example add support for CSV file, simply adding a decoder, but then the code will throw an exception because the `text/csv` type isn't included in the list, so we need to provide also a custom implementation of **IMimeTypeDetection**.

We should release the changes to Abstractions first, then it's up to you, though I think it would be nice shipping the work done so far.

Could you move these changes to a separate PR, so we can merge them and release the nuget?

image

service/Abstractions/Pipeline/DataPipeline.cs
```diff
@@ -149,10 +149,7 @@ public void MarkProcessedBy(IPipelineStepHandler handler)
         /// <param name="text">Text to store for the end user</param>
         public void Log(IPipelineStepHandler handler, string text)
         {
-            if (this.LogEntries == null)
-            {
-                this.LogEntries = new List<PipelineLogEntry>();
-            }
+            this.LogEntries ??= new List<PipelineLogEntry>();
```
dluc, line 152:
```suggestion
            if (this.LogEntries == null)
            {
                this.LogEntries = new List<PipelineLogEntry>();
            }
```

reverting changes to reduce size of Abstractions changes

dluc:
the new Abstractions nuget is out, you should be able to upgrade in a few minutes and use the new code. Version: 0.36.240412.1

marcominerva:
Thank you @dluc, I have updated the PR using the new **Abstractions** NuGet package.

dluc (changes requested):
Please check all the decoders, looks like they are not setting FileContent.MimeType before returning the result

marcominerva:
The `FileContent.MimeType` property has the value `MimeTypes.PlainText` by default, so I think there is no need to set it explicitly, expect in the `MarkDownDecoder`. Is it right?

dluc (approved):

PR #399, opened 2024-04-10
Opinionated Deployment Scripts

cherchyk:


dluc:
When you have a chance, could you squash the changes into a single commit? The PR will be easier to manage in case we have other changes and we need to rebase. Thanks

cherchyk:
> When you have a chance, could you squash the changes into a single commit? The PR will be easier to manage in case we have other changes and we need to rebase. Thanks

this is the best I can do.  I merged with the Main somewhere in the middle and now I cannot squash commits before this merge

Issue #100, opened 2023-10-18
Streaming AskAsync response


cz35iek:
Hi,

We're using your library in our project. It get us up to speed preety fast in these new AI topics, so thank you for that :). 

I was wondering if there is an option to recieve streaming response from our semantic memory service. This could improve user experience, to recieve some first tokens of response instead of waiting for whole thing. OpenAI chatgpt is working this way. 

BR,
Dawid

dluc:
+1 that's on the list, I agree it's highly needed for responsive UI. We don't have a timeline yet, but it's high on the priority list

JohnGalt1717:
Here here. 

Since virtually all of the LLMs do this by default it would be nice to just add AskStreamingAsync and return IAsyncEnumerable.  I think just one method needs to be added to the ITextGenerator and then the included ones updated.

roldengarm:
@dluc  we also need this. Right now, after ingesting ~125k documents, and pausing all ingestion, it still takes about 10-15 seconds to get an answer. This is too long for our requirements; users expect responses within seconds. 
Any ETA on the streaming response? And, any suggestions how we can improve performance? Based on the App Insights dependency log, the bottleneck seems to be the embeddings:

![image](https://github.com/microsoft/kernel-memory/assets/37638588/c39c2274-cd0e-4afd-ad55-8ebb1136ad10)
![image](https://github.com/microsoft/kernel-memory/assets/37638588/e86c1fcb-7d61-4c8a-a91b-3ba9bee69576)

Is there any setting we can tweak to improve performance?
Thanks.

roldengarm:
quick update; the main bottleneck for us is the LLM, with GPT35 it's a lot faster. But, we need GPT4, so streaming is essential.

@dluc  I may be able to assist for the implementation.
I've analysed the solution and my proposal would be the following, of course open to your feedback :)

- Add another method AskStreamingAsync based on AskAsync [here](https://github.com/microsoft/kernel-memory/blob/main/service/Core/Search/SearchClient.cs#L202)
- The results to stream are currently generated [here](https://github.com/microsoft/kernel-memory/blob/main/service/Core/Search/SearchClient.cs#L355), so that could be replaced with e.g. a yield return.

For the API endpoint
- End point to be added [here](https://github.com/microsoft/kernel-memory/blob/main/service/Service/Program.cs#L266)
- This should call AskStreamingAsync in the SearchClient
[...]

PR #400, opened 2024-04-12
Implement new streaming ask endpoint (WIP)

JonathanVelkeneers:
Draft PR for issue https://github.com/microsoft/kernel-memory/issues/100 .
This draft pr was very briefly discussed in discord. The changes still need work.
currently missing:
- more examples (serverless, service, curl,...)
- ~~Get the streaming working in WebClient, it currently still buffers the response (i was not able to find a fix)~~

Used existing AskAsync method as reference point for most of the code.
The new streaming endpoint does not return a MemoryAnswer, but returns the actual text result in an async enumerable.
If sources are needed the SearchAsync method can still be used to fetch them with minimal extra performance overhead.


JonathanVelkeneers:

clients/dotnet/WebClient/MemoryWebClient.cs
```diff
@@ -299,6 +300,52 @@ public async Task DeleteDocumentAsync(string documentId, string? index = null, C
...
+            Question = question,
+            Filters = (filters is { Count: > 0 }) ? filters.ToList() : new(),
+            MinRelevance = minRelevance
+        };
+        using StringContent content = new(JsonSerializer.Serialize(request), Encoding.UTF8, "application/json");
+
+        using var httpRequest = new HttpRequestMessage(HttpMethod.Post, Constants.HttpAskStreamEndpoint);
+        httpRequest.Content = content;
+
+        HttpResponseMessage response = await this._client.SendAsync(httpRequest, cancellationToken).ConfigureAwait(false);
```
JonathanVelkeneers, line 331: This call buffers the whole response for some reason. I can't figure out why.
- [ ]  Fix buffering

JonathanVelkeneers:
Also I want to note, the way I have this implemented currently is just returning the whole answer as a stream of strings.
This might not be optimal as the code that consumes these methods needs to call SearchAsync() if they want the sources, and this would result in an extra DB query.

Another implementation i was thinking of was returning a stream of PartialMemoryAnswer(MemoryAnswer with nullable properties), with the first object returning the metadata (such as facts), and the parts after the first one only containing their pieces of text.
This would also mean that the object can be extended upon in the future.

example:

``` c#

var noAnswerFound = new PartialMemoryAnswer
{
    Question = question,
    NoResult = true,
    Result = this._config.EmptyAnswer,
};

StringBuilder bufferedAnswer = new(); // Buffering result in memory to count chars and check if not empty answer
bool finishedRequiredBuffering = false;
var watch = Stopwatch.StartNew();
await foreach (var x in this.GenerateAnswerAsync(question, facts.ToString()).WithCancellation(cancellationToken).ConfigureAwait(false))
{
    if (x is null || x.Length == 0)
    {
        continue;
    }

    bufferedAnswer.Append(x);
    int currentLength = bufferedAnswer.Length;

    if (!finishedRequiredBuffering)
    {
        if (currentLength <= this._config.EmptyAnswer.Length && ValueIsEquivalentTo(bufferedAnswer.ToString(), this._config.EmptyAnswer))
        {
            this._log.LogTrace("Answer generated in {0} msecs. No relevant memories found", watch.ElapsedMilliseconds);
            noAnswerFound.NoResultReason = "No relevant memories found";
            yield return noAnswerFound;
            yield break;
        }
        else if (currentLength > this._config.EmptyAnswer.Length)
        {
            finishedRequiredBuffering = true;
            yield return new PartialMemoryAnswer
            {
                NoResult = false,
                Result = bufferedAnswer.ToString(),
                RelevantSources = citationList,
            };
        }
    }

    if (finishedRequiredBuffering)
    {
        yield return new PartialMemoryAnswer
        {
            Result = x
        };
    }

    if (this._log.IsEnabled(LogLevel.Trace) && currentLength >= 30)
    {
        this._log.LogTrace("{0} chars generated", currentLength);
    }
}

```

clients/dotnet/WebClient/MemoryWebClient.cs
JonathanVelkeneers: Thanks! @marcominerva. I was able to fix it with those changes. 

The response in the webclient when printed is (these are the exact characters, not a stringified list):
```json
["Yes",","," NASA"," has"," invited"," media"," to"," see"," the"," new"," test"," version"," of"," the"," Orion"," spacecraft"," and"," the"," hardware"," teams"," will"," use"," to"," recover"," the"," capsule"," and"," astronauts"," upon"," their"," return"," from"," space"," during"," the"," Artem","is"," II"," mission","."," The"," event"," will"," take"," place"," at"," Naval"," Base"," San"," Diego"," on"," August"," ","2","."," Personnel"," involved"," in"," recovery"," operations"," from"," NASA",","," the"," U",".S","."," Navy",","," and"," the"," U",".S","."," Air"," Force"," will"," be"," available"," to"," speak"," with"," media","."," Teams"," are"," currently"," conducting"," tests"," in"," the"," Pacific"," Ocean"," to"," demonstrate"," and"," evaluate"," the"," processes",","," procedures",","," and"," hardware"," for"," recovery"," operations"," for"," crew","ed"," Artem","is"," missions","."," The"," tests"," will"," help"," prepare"," the"," team"," for"," Artem","is"," II",","," NASA","'s"," first"," crew","ed"," mission"," under"," Artem","is"," that"," will"," send"," four"," astronauts"," in"," Orion"," around"," the"," Moon"," to"," checkout"," systems"," ahead"," of"," future"," lunar"," missions","."," The"," Artem","is"," II"," crew"," will"," participate"," in"," recovery"," testing"," at"," sea"," next"," year","."]
```
Which is not very clean, and is not really usable.

If i swap the whole codeblock for .NET8.0's ReadFromJsonAsAsyncEnumerable<>() method it works and give a clean string as response, but that would mean migrating the WebClient project to .NET8.0.

marcominerva: The `ReadFromJsonAsAsyncEnumerable<>()` method should be available in the https://www.nuget.org/packages/System.Net.Http.Json NuGet package, that is compatible also for .NET Standard 2.0. You can try adding this package to the project.

JonathanVelkeneers: Great point! Added the dependency and it works!
The output in the webclient example prints as it is supposed to now.

service/Core/MemoryServerless.cs
```diff
@@ -257,4 +257,29 @@ public Task DeleteDocumentAsync(string documentId, string? index = null, Cancell
             minRelevance: minRelevance,
             cancellationToken: cancellationToken);
     }
+
+    /// <inheritdoc />
+    public IAsyncEnumerable<string> AskStreamingAsync(
```
dluc, line 262: this new method returns only the answer, missing information about sources, relevance, etc. Why not stream a json object?

JonathanVelkeneers: > this new method returns only the answer, missing information about sources, relevance, etc. Why not stream a json object?

I did a small write-up [here](https://github.com/microsoft/kernel-memory/pull/400#issuecomment-2075310644).

In this PR I made it as a stream of strings, because using an object would require making some design choices.
Is the extra information (sources, relevance, etc) sent with every "answer part", or only the first/last one?

In an implementation I did in a personal project I sent all the metadata in the first response enumerable entry, and only sent messageID + messageChunk with every remaining part.

If you prefer that approach or have another solution I can rewrite these changes.

dluc: I think everyone will want streaming, without giving up the other metadata, in particular the list of sources and their relevance. I would use the same approach used by OpenAI streaming, returning a content_update property within the response object, that includes the token - potentially sending the list of sources on the first response only using the same approach (something like sources_update) to avoid the overhead of sending the same sources with every token.

Something like, pseudo structure (I would reuse the existing class, just change it a bit to support "_update"):

response: 
```
{
    sources_update {
          stream of items.   // sent all at once with the first token, then the list stays empty
   },
   content_update {
          next string token
   }
}
```

dluc:
some conflicts to address

chaelli:
> some conflicts to address

if it helps - I rebased and fixed some issues => https://github.com/chaelli/kernel-memory/tree/100-streaming-askAsync-response
not sure how to handle this. PR into @JonathanVelkeneers' branch? Or create a separat PR so that the validation tasks can run?
Anyway... probably should do some more testing before that.

JonathanVelkeneers:
> some conflicts to address

Resolved these

chaelli:
any reason this is waiting? do we need to add SSE format for the endpoint? (if so - maybe this works https://github.com/microsoft/kernel-memory/discussions/625#discussioncomment-9698298 - did for my quick test in the browser - but my experience with SSE are minimal ;))

dluc:
> @dluc just checking in, can this be merged soon please? We're about to build a UI using Kernel Memory and streaming is a must-have
> 
> @JonathanVelkeneers can you please check as well? Thanks heaps, would be so great to have streaming capabilities, and I believe we're not the only one :)

sorry to keep this on hold but I cannot merge yet. I'd suggest to work with a fork not to be blocked. It's an important feature but there's some other critical work in progress.

chaelli:
@JonathanVelkeneers I played around a bit with SSE and now have a working solution that does "clean" SSE (via get). Basically, I use two endpoints - one that starts the answer generation and returns an id and one that then streams the answer:


```private static readonly Dictionary<string, IAsyncEnumerable<MemoryAnswer>> s_askStreams = new();

public static void AddAskStreamEndpoint(
        this IEndpointRouteBuilder builder, string apiPrefix = "/", IEndpointFilter? authFilter = null)
    {
        RouteGroupBuilder group = builder.MapGroup(apiPrefix);

        // Ask streaming endpoint
        var route = group.MapPost(Constants.HttpAskStreamEndpoint, async Task<IResult> (
                HttpContext context,
                MemoryQuery query,
                IKernelMemory service,
                ILogger<KernelMemoryWebAPI> log,
                CancellationToken cancellationToken) =>
            {
                log.LogTrace("New search request, index '{0}', minRelevance {1}", query.Index, query.MinRelevance);

                var askId = Guid.NewGuid().ToString("N").Substring(0, 8);
                s_askStreams.Add(askId, service.AskStreamingAsync(
                    question: query.Question,
                    index: query.Index,
                    filters: query.Filters,
                    minRelevance: query.MinRelevance,
                    cancellationToken: cancellationToken));

                return Results.Ok(new AskStreamResponse { AskId = askId });
            })
            .Produces<AskStreamResponse>(StatusCodes.Status200OK)
            .Produces<ProblemDetails>(StatusCodes.Status401Unauthorized)
            .Produces<ProblemDetails>(StatusCodes.Status403Forbidden);

        if (authFilter != null) { route.AddEndpointFilter(authFilter); }

        // SSE endpoint
        var sseRoute = group.MapGet($"{Constants.HttpAskStreamEndpoint}/{{askId}}", async Task (
            HttpContext context,
            ILogger<KernelMemoryWebAPI> log,
            CancellationToken cancellationToken,
            string askId) =>
            {
                log.LogTrace("Accessing SSE stream for askId '{0}'", askId);
                context.Response.Headers.Add("Content-Type", "text/event-stream");
                if (s_askStreams.TryGetValue(askId, out IAsyncEnumerable<MemoryAnswer>? stream))
                {
                    s_askStreams.Remove(askId);
                    var jsonOptions = new JsonSerializerOptions { DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingDefault };
                    var response = context.Response;
                    if (stream != null)
                    {
                        await foreach (var ma in stream)
                        {
                            await response.WriteAsync($"data: {JsonSerializer.Serialize(new { message = ma }, jsonOptions)}\n\n", cancellationToken).ConfigureAwait(false);
                            await response.Body.FlushAsync(cancellationToken).ConfigureAwait(false);
                        }

                        // Send an end-of-stream event if needed
                        await response.WriteAsync("event: end\ndata: End of stream\n\n", cancellationToken).ConfigureAwait(false);
                        await response.Body.FlushAsync(cancellationToken).ConfigureAwait(false);
                        return;
                    }
                }
                context.Response.StatusCode = StatusCodes.Status404NotFound;
                await context.Response.WriteAsync("Not found", cancellationToken).ConfigureAwait(false);
            });```

When requesting the answer from the browser, this is the basic functionality:
```var response = await fetch("http://localhost:9001/ask/stream", {
  "headers": {
    "accept": "*/*",
    "authorization": "...",
    "cache-control": "no-cache",
    "Content-Type": "application/json"
  },
  "body": "{\"question\":\"was macht viu?\",\"index\":\"viuch-largeembedding\"}",
  "method": "POST",
  "mode": "cors",
  "credentials": "include"
});
const responseData = await response.json();
const askId = responseData.askId;

const sseUrl = `http://localhost:9001/ask/stream/${askId}`;
const eventSource = new EventSource(sseUrl);

let responseText = '';
eventSource.onmessage = function(event) {
    const responeObject = JSON.parse(event.data);
    responseText += responeObject.message.text;
    console.log(responseText);
};

eventSource.addEventListener('end', function(event) {
    console.log('End of stream');
    eventSource.close(); // Close the SSE connection
});```

maybe you can add this to your branch?

chaelli:
@dluc does it make sense for us to rebase this? or will it take much longer to merge? 
And @JonathanVelkeneers  if you do rebase (or merge & fix) - I'd suggest you extract the fact generation & token calculation into a separate method - otherwise we'll always need to fix this stuff twice if it changes.
I can certainly help - but I'd rather only invest the time once we know theres a good chance it will be merged

dluc:
> @dluc does it make sense for us to rebase this? or will it take much longer to merge? And @JonathanVelkeneers if you do rebase (or merge & fix) - I'd suggest you extract the fact generation & token calculation into a separate method - otherwise we'll always need to fix this stuff twice if it changes. I can certainly help - but I'd rather only invest the time once we know theres a good chance it will be merged

no worries I'll take care of the rebase ;-)

stavroskasidis:
Is this feature still gonna be implemented?

dluc:
> Is this feature still gonna be implemented?

Apologies for the delay, I had to put this topic on hold, and we currently have two PRs for the same feature. I’ll need to review and decide the best approach. Unfortunately, no ETA at the moment

dluc:
Update: for this feature to be merged, there's a couple of things to do:
* Check this similar PR https://github.com/microsoft/kernel-memory/pull/726 and decide which approach to take
* Support content moderation. The stream of tokens needs to be validated while streamed, on a configurable frequence. If at any point the text moderation fails, the stream needs to be reset, e.g. sending a special token or similar.

nurkmez2:
@JonathanVelkeneers 
It would be great if you could complete this feature.
Thanks

dluc:
@JonathanVelkeneers thanks for helping! - There were two PRs addressing the same feature, and I had to decide between them. The Response Streaming implementation is now merged into `main`. For more details, please refer to PR #726.

PR #407, opened 2024-04-16
Allow to inject custom web scraper

dluc:


JonathanVelkeneers:
Looks good!
Could maybe use an example in the example folder so it's a bit documented.

dluc:
> Looks good! Could maybe use an example in the example folder so it's a bit documented.

example added 👍

PR #409, opened 2024-04-17
Bicep optimizations

cherchyk:

### DIFF
--- a/examples/001-dotnet-WebClient/Program.cs
+++ b/examples/001-dotnet-WebClient/Program.cs
@@ -23,7 +23,7 @@ public static class Program
     private static readonly List<string> s_toDelete = new();
 
     // Change this to True and configure Azure Document Intelligence to test OCR and support for images
-    private const bool imageSupportDemoEnabled = true;
+    private const bool ImageSupportDemoEnabled = false;
 
     public static async Task Main()
     {
@@ -100,7 +100,7 @@ public static class Program
     // Extract memory from images (OCR required)
     private static async Task StoreImage()
     {
-        if (!imageSupportDemoEnabled) { return; }
+        if (!ImageSupportDemoEnabled) { return; }
 
         Console.WriteLine("Uploading Image file with a news about a conference sponsored by Microsoft");
         await s_memory.ImportDocumentAsync(new Document("img001").AddFiles(new[] { "file6-ANWC-image.jpg" }));
@@ -314,7 +314,7 @@ public static class Program
 
         var answer = await s_memory.AskAsync(question, minRelevance: 0.76);
 
-        Console.WriteLine(imageSupportDemoEnabled
+        Console.WriteLine(ImageSupportDemoEnabled
             ? $"\nAnswer: {answer.Result}\n\n  Sources:\n"
             : $"\nAnswer (none expected): {answer.Result}\n\n  Sources:\n");
 
--- a/examples/001-dotnet-WebClient/Program.cs
+++ b/examples/001-dotnet-WebClient/Program.cs
@@ -23,7 +23,7 @@ public static class Program
     private static readonly List<string> s_toDelete = new();
 
     // Change this to True and configure Azure Document Intelligence to test OCR and support for images
-    private const bool ImageSupportDemoEnabled = false;
+    private const bool ImageSupportDemoEnabled = true;
 
     public static async Task Main()
     {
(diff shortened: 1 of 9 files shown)


examples/001-dotnet-WebClient/Program.cs
```diff
@@ -23,7 +23,7 @@ public static class Program
     private static readonly List<string> s_toDelete = new();
 
     // Change this to True and configure Azure Document Intelligence to test OCR and support for images
-    private const bool imageSupportDemoEnabled = true;
```
dluc, line 26: could you update the bicep template to include Azure Document Intelligence? that service is really useful and we're going to use it more. I rather leave this enabled for now.

infra/main.bicep
```diff
@@ -21,6 +21,20 @@ For model limits specific to your region, refer to the documentation at https://
 @maxValue(40)
 param embeddingDeploymentCapacity int = 30
 
+@description('''
+Kernel Memory Service Authorization AccessKey1. It will be passed to the Kernel Memory Service container as an environment variable.
```
dluc, line 25:
```suggestion
Kernel Memory Service Authorization AccessKey1. 
It's passed to the Kernel Memory Service container as an environment variable
and is required to authenticate HTTP requests to KM web service.
```

infra/main.bicep
```diff
@@ -21,6 +21,20 @@ For model limits specific to your region, refer to the documentation at https://
...
+''')
+@minLength(16)
+@maxLength(42)
+param KernelMemoryServiceAuthorizationAccessKey1 string = 'KernelMemoryServiceAuthorizationAccessKey1'
+
+@description('''
+Kernel Memory Service Authorization AccessKey2. It will be passed to the Kernel Memory Service container as an environment variable.
```
dluc, line 32:
```suggestion
Kernel Memory Service Authorization AccessKey2.
It's passed to the Kernel Memory Service container as an environment variable
and is required to authenticate HTTP requests to KM web service.
```

infra/main.json
```diff
@@ -4,8 +4,8 @@
   "metadata": {
     "_generator": {
       "name": "bicep",
-      "version": "0.26.170.59819",
-      "templateHash": "12294581750482242115"
+      "version": "0.23.1.45101",
```
dluc, line 7: nit: you might want to update your AZ CLI, that might also help fixing the various warnings thrown when compiling to JSON

infra/main.bicep
```diff
@@ -21,6 +21,20 @@ For model limits specific to your region, refer to the documentation at https://
 @maxValue(40)
 param embeddingDeploymentCapacity int = 30
 
+@description('''
+Kernel Memory Service Authorization AccessKey1. It will be passed to the Kernel Memory Service container as an environment variable.
+''')
+@minLength(16)
+@maxLength(42)
+param KernelMemoryServiceAuthorizationAccessKey1 string = 'KernelMemoryServiceAuthorizationAccessKey1'
```
dluc, line 29: I would autogenerate a random value similarly to suffix. The user can still change it, but at least it's safe to begin with

cherchyk: in the UI it will look like this  `[substring(newGuid(), 0, 6)]`

examples/001-dotnet-WebClient/Program.cs
cherchyk: let me do this as a next PR

infra/main.json
cherchyk: I was building with `bicep build main.bicep`
you are using `az bicep build --file main.bicep`

I removed all warnings except one.  It was not related to compiler but how I had bicep written.

infra/main.bicep
dluc: I'm not sure I understand... 
`suffix` is defined as `substring(newGuid(), 0, 6)` while `KernelMemoryServiceAuthorizationAccessKey1` and `KernelMemoryServiceAuthorizationAccessKey2` are defined as constant values:

```
param suffix string = substring(newGuid(), 0, 6)

param KernelMemoryServiceAuthorizationAccessKey1 string = 'KernelMemoryServiceAuthorizationAccessKey1'

param KernelMemoryServiceAuthorizationAccessKey2 string = 'KernelMemoryServiceAuthorizationAccessKey2'
```

shouldn't these two also use `substring` (and longer than 6 chars) ?

examples/001-dotnet-WebClient/Program.cs
dluc: let's leave this change out for now

infra/main.bicep
dluc: I think the best approach is generating a random value, so the deployment is immediately secure - using the same approach used for suffix, just longer than 6 chars

infra/main.json
dluc: yeah I didn't want to install yet another tool, it seems counterintuitive considering how much effort has been put into `az`

dluc (approved):

PR #415, opened 2024-04-18
Export/Download document support

coryisakson:

Validating AI answers requires access to the source grounding documents and data. The KM solution enables easy ingestion of grounding documents as well as the ability to remove documents. A file download feature allows consumers access to the grounding source materials and allow them to verify the answers presented by the ASK endpoint.

* New methods to memory and storage interface, to allow access to individual files
* New web service endpoint to download files, passing index name, document ID and file name
* Include download link in RAG citations and search results:
image
* Stream file download and support Range fetch

Since KM is a backend service not meant for multi-user direct access (ie KM security model is based on a single key, like a SQL server or any DB), the endpoint provides direct access to all files, similarly to search allows access to all memory records. For public deployments, a middleware webservice should take care of securing links, e.g. adding and validating signatures and user tokens.

### DIFF
--- a/clients/dotnet/WebClient/MemoryWebClient.cs
+++ b/clients/dotnet/WebClient/MemoryWebClient.cs
@@ -6,6 +6,7 @@ using System.IO;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
+using System.Net.Http.Headers;
 using System.Text;
 using System.Text.Json;
 using System.Threading;
@@ -233,12 +234,42 @@ public class MemoryWebClient : IKernelMemory
         return status;
     }
 
-#if KernelMemoryDev
-    public Task<StreamableFileContent> ExportFileAsync(string documentId, string fileName, string? index = null, CancellationToken cancellationToken = default)
+    /// <inheritdoc />
+    public async Task<StreamableFileContent> ExportFileAsync(
+        string documentId,
+        string fileName,
+        string? index = null,
+        CancellationToken cancellationToken = default)
     {
-        throw new NotImplementedException();
+        string indexParam = index != null ? $"index={index}&" : "";
+        string requestUri = $"{Constants.HttpDownloadEndpoint}?{indexParam}documentId ={documentId}&filename={fileName}";
+        HttpResponseMessage? response = await this._client.GetAsync(requestUri, cancellationToken).ConfigureAwait(false);
+
+        response.EnsureSuccessStatusCode();
+
+        // Header example: Content-Type: text/csv
+        response.Headers.TryGetValues("Content-Type", out IEnumerable<string> contentTypeValues);
+        string contentType = (from x in contentTypeValues select x).FirstOrDefault() ?? "application/octet-stream";
+
+        // Header example: Content-Disposition: attachment; filename=data.csv; filename*=UTF-8''data.csv
+        response.Headers.TryGetValues("Content-Disposition", out IEnumerable<string> contentDispositionValues);
+        var disposition = ContentDispositionHeaderValue.Parse(contentDispositionValues.FirstOrDefault());
+
+        // Header example: Last-Modified: Wed, 01 May 2024 04:56:10 GMT
+        response.Headers.TryGetValues("Last-Modified", out IEnumerable<string> lastModifiedValues);
+        if (!DateTimeOffset.TryParse((from x in lastModifiedValues select x).FirstOrDefault(), out DateTimeOffset lastModifiedValue))
+        {
+            lastModifiedValue = DateTimeOffset.MinValue;
+        }
+
+        StreamableFileContent result = new(
+            fileName: disposition.FileName ?? "file.bin",
+            fileSize: disposition.Size.GetValueOrDefault(),
+            fileType: contentType,
+            lastWriteTimeUtc: lastModifiedValue,
+            asyncStreamDelegate: response.Content.ReadAsStreamAsync);
+        return result;
     }
-#endif
 
     /// <inheritdoc />
     public async Task<SearchResult> SearchAsync(
--- a/service/Abstractions/Constants.cs
+++ b/service/Abstractions/Constants.cs
@@ -61,8 +61,10 @@ public static class Constants
     public const string HttpDeleteDocumentEndpointWithParams = $"{HttpDocumentsEndpoint}?{WebServiceIndexField}={HttpIndexPlaceholder}&{WebServiceDocumentIdField}={HttpDocumentIdPlaceholder}";
     public const string HttpDeleteIndexEndpointWithParams = $"{HttpIndexesEndpoint}?{WebServiceIndexField}={HttpIndexPlaceholder}";
     public const string HttpUploadStatusEndpointWithParams = $"{HttpUploadStatusEndpoint}?{WebServiceIndexField}={HttpIndexPlaceholder}&{WebServiceDocumentIdField}={HttpDocumentIdPlaceholder}";
+    public const string HttpDownloadEndpointWithParams = $"{HttpDownloadEndpoint}?{WebServiceIndexField}={HttpIndexPlaceholder}&{WebServiceDocumentIdField}={HttpDocumentIdPlaceholder}&{WebServiceFilenameField}={HttpFilenamePlaceholder}";
     public const string HttpIndexPlaceholder = "{index}";
     public const string HttpDocumentIdPlaceholder = "{documentId}";
+    public const string HttpFilenamePlaceholder = "{filename}";
 
     // Pipeline Handlers, Step names
     public const string PipelineStepsExtract = "extract";
--- a/service/Service.AspNetCore/WebAPIEndpoints.cs
+++ b/service/Service.AspNetCore/WebAPIEndpoints.cs
@@ -1,5 +1,9 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
+// ReSharper disable RedundantUsingDirective
+
+#pragma warning disable IDE0005 // temp
+
 using System;
 using System.Collections.Generic;
 using System.Threading;
@@ -9,7 +13,10 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.Logging;
+using Microsoft.KernelMemory.ContentStorage;
 using Microsoft.KernelMemory.Service.AspNetCore.Models;
+using System.IO;
+using Microsoft.AspNetCore.Http.HttpResults;
 
 namespace Microsoft.KernelMemory.Service.AspNetCore;
 
@@ -27,6 +34,9 @@ public static class WebAPIEndpoints
         builder.AddAskEndpoint(apiPrefix, authFilter);
         builder.AddSearchEndpoint(apiPrefix, authFilter);
         builder.AddUploadStatusEndpoint(apiPrefix, authFilter);
+#if KernelMemoryDev
+        builder.AddGetDownloadEndpoint(apiPrefix, authFilter);
+#endif
 
         return builder;
     }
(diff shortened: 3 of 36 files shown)


coryisakson:
@microsoft-github-policy-service agree [company="Microsoft"]

coryisakson:
@microsoft-github-policy-service agree company="Microsoft"

dluc:
@coryisakson could you merge the latest changes from main and check che build? I can't build the code, with a bunch or warnings and errors. Thanks

coryisakson:
@dluc the branch is updated and unit tests are passing.

dluc:
I tried fixing the conflicts and reviewing but the PR it too big. I think it would really help excluding changes that are unrelated to the new feature, e.g. spacing, string changes (like the mime type in qdrant). E.g. rather than 52 files maybe bring the PR down to 30 files or so. Given the big number of changed to interfaces and new classes it's going to take some time. 
By the way I tried to rebuild the branch so that it can be more easily rebased when main changes, but even that was taking too long. If you have a chance to squash all the changes into a single commit, that would help managing the PR.

.vscode/launch.json
```diff
@@ -0,0 +1,7 @@
+{
```
dluc, line 1: nit: changes out of scope, this is one file I would leave out of this PR to reduce the PR size (it can be added in a separate PR if it helps)

.vscode/settings.json
```diff
@@ -20,7 +20,8 @@
     "**/.hg": true,
```
dluc, line 20: nit: changes out of scope, this is one file I would leave out of this PR to reduce the PR size (it can be added in a separate PR if it helps)

extensions/Qdrant/Qdrant/Client/Http/HttpRequest.cs
```diff
@@ -5,6 +5,7 @@
 using System.Net.Http.Headers;
```
dluc, line 5: nit: changes out of scope, this is one file I would leave out of this PR

nuget.config
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
dluc, line 1: nit: changes out of scope, this is one file I would leave out of this PR

service/Abstractions/Compositions/ExportValidationService.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using System.Threading;
+using Microsoft.KernelMemory.ContentStorage;
+using Microsoft.KernelMemory.Search;
+using System.Linq;
+
+namespace Microsoft.KernelMemory.Compositions;
+
+public class ExportValidationService : IExportValidationService
```
dluc, line 12: I would move this class to Core, trying to avoid having (too much) logic in the Abstractions assembly

service/Abstractions/Constants.cs
```diff
@@ -95,4 +96,9 @@ public static class Constants
     // Standard prompt names
     public const string PromptNamesSummarize = "summarize";
     public const string PromptNamesAnswerWithFacts = "answer-with-facts";
+
+    // File Download Header
```
dluc, line 100: why are these headers needed?

service/Service/Service.csproj
```diff
@@ -9,37 +9,38 @@
         <ManagePackageVersionsCentrally>false</ManagePackageVersionsCentrally>
```
dluc, line 9: nit: a lot of changes out of scope here, I don't think any are needed

service/tests/TestHelpers/TestHelpers.csproj
```diff
@@ -9,6 +9,7 @@
         <IsPackable>false</IsPackable>
         <GenerateDocumentationFile>false</GenerateDocumentationFile>
         <NoWarn>$(NoWarn);CS1591;CA1861;CA1051;</NoWarn>
+        <UserSecretsId>fd097735-4bda-4320-ad57-54ca2fea0e5d</UserSecretsId>
```
dluc, line 12: this is not needed

service/Abstractions/ContentStorage/IContentStorage.cs
```diff
@@ -66,13 +66,15 @@ public interface IContentStorage
...
     /// <param name="fileName">Name of the file</param>
+    /// <param name="contentType">Mime type of the file</param>
     /// <param name="streamContent">File content</param>
     /// <param name="cancellationToken">Async task cancellation token</param>
     Task WriteFileAsync(
         string index,
         string documentId,
         string fileName,
         Stream streamContent,
+        string contentType = "application/octet-stream",
```
dluc, line 77: I would move this change to a separate PR to be reviewed and merged first. By the way, any change to Abstractions needs to be done separately due to how nugets are wired up.
Something like this:

PR1: changes to IContentStorage
PR2: changes to classes implementing IContentStorage
PR3: rest of the PR (we'll have to split again probably but for now we can do all in one smaller PR)

service/Abstractions/ContentStorage/IContentStorage.cs
```diff
@@ -90,4 +92,20 @@ public interface IContentStorage
...
+    /// <summary>
+    /// Fetch file info from storage
+    /// </summary>
+    /// <param name="index">Index name</param>
+    /// <param name="documentId">Document ID</param>
+    /// <param name="fileName"></param>
+    /// <param name="logErrIfNotFound">Whether to log an error if the file does not exist. An exception will be raised anyway.</param>
+    /// <param name="cancellationToken">Async task cancellation token</param>
+    /// <returns>File content</returns>
+    Task<IContentFile> FileInfoAsync(
```
dluc, line 105: 1. rename to `GetFileInfoAsync`
2. why do we need this new method?
3. why return IContentFile (an interface) rather than a specific type?

service/Abstractions/ContentStorage/IContentFile.cs
```diff
@@ -0,0 +1,57 @@
...
+    string? Volume { get; }
+    string? RelativePath { get; }
+    string? FileName { get; }
+    string? ContentType { get; }
+    DateTimeOffset? LastWrite { get; }
+    Func<Task<Stream>> StreamAsync { get; }
+    long Length { get; }
+}
+
+public sealed class StreamableContentFile : IContentFile, IDisposable
```
dluc, line 19: move to separate file

why do we need an interface for this class?

service/Abstractions/ContentStorage/IContentFile.cs
```diff
@@ -0,0 +1,57 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Microsoft.KernelMemory.ContentStorage;
+public interface IContentFile
+{
```
dluc, line 9: can we simplify here, do we need all these properties?

dluc (changes requested):

service/Abstractions/ContentStorage/IContentFile.cs
coryisakson: Interface is not required, I can remove it and just use the Concrete class

nuget.config
```diff
@@ -1,6 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
-
-<configuration>
+﻿<configuration>
```
dluc, line 1:
```suggestion
<?xml version="1.0" encoding="utf-8"?>

<configuration>
```

clients/dotnet/SemanticKernelPlugin/MemoryPlugin.cs
```diff
@@ -67,7 +67,7 @@ public class MemoryPlugin
     /// <summary>
     /// Name of the input variable used to specify optional tags associated with stored information.
     ///
-    /// Tags can be used to filter memories over one or multiple keys, e.g. userID, tenant, groups,
+    /// Tags can be used to Filter memories over one or multiple keys, e.g. userID, tenant, groups,
```
dluc, line 70:
```suggestion
    /// Tags can be used to filter memories over one or multiple keys, e.g. userID, tenant, groups,
```

clients/dotnet/SemanticKernelPlugin/MemoryPlugin.cs
```diff
@@ -101,7 +101,7 @@ public class MemoryPlugin
     private readonly TagCollection? _defaultIngestionTags = null;
 
     /// <summary>
-    /// Default collection of tags required when retrieving memory (using filters).
+    /// Default collection of tags required when retrieving memory (using Filters).
```
dluc, line 104:
```suggestion
    /// Default collection of tags required when retrieving memory (using filters).
```

examples/001-dotnet-WebClient/Program.cs
```diff
@@ -233,7 +233,7 @@ private static async Task StoreJson()
     // === RETRIEVAL =========
     // =======================
 
-    // Question without filters
+    // Question without Filters
```
dluc, line 236: why all these "Filter" changes? :-)

extensions/AzureAISearch/AzureAISearch/AzureAISearchFiltering.cs
```diff
@@ -13,22 +13,22 @@ internal static class AzureAISearchFiltering
...
-    /// * One filter can have multiple key-values: these are combined with AND conditions
-    /// * Multiple filters are combined using OR
-    /// * Multiple filters can apply to the same key:
-    ///     * if filters have only one condition, these are rendered using "search.in"
-    ///     * if filters have multiple conditions, the values are combined with OR
+    /// * One Filter can have multiple key-values: these are combined with AND conditions
+    /// * Multiple Filters are combined using OR
+    /// * Multiple Filters can apply to the same key:
+    ///     * if Filters have only one condition, these are rendered using "search.in"
+    ///     * if Filters have multiple conditions, the values are combined with OR
```
dluc, line 20: looks like some find&replace went wild here :-) could you undo all the "filter" to "Filter" changes?

examples/001-dotnet-WebClient/Program.cs
dluc: fixed

extensions/AzureAISearch/AzureAISearch/AzureAISearchFiltering.cs
dluc: fixed

service/tests/TestHelpers/TestHelpers.csproj
dluc: fixed

dluc:
I rebuilt the branch fixing some merge gone wrong, and making a few minor changes to namespaces and names. I see the approach taken introduces a new dependency with the responsibility of checking access and downloading files, which I'm not sure about, in terms of design. I'll try playing with some changes, reorganizing these responsibilities and how memory/storage/orchestrator work together to provide the same functionality. My preference would be about leveraging the orchestrator, not nesting content access into the validation service (which should just validate). I haven't looked at the download part yet, e.g IContent interface, which might actually be more important given it affects the primary API.

service/Service.AspNetCore/WebAPIEndpoints.cs
```diff
@@ -296,6 +299,78 @@ await service.DeleteDocumentAsync(documentId: documentId, index: index, cancella
         if (authFilter != null) { route.AddEndpointFilter(authFilter); }
     }
 
+    public static void AddPostDownloadEndpoint(this IEndpointRouteBuilder builder, string apiPrefix = "/", IEndpointFilter? authFilter = null)
+    {
+        RouteGroupBuilder group = builder.MapGroup(apiPrefix);
+
+        // File download endpoint
+        var route = group.MapPost(Constants.HttpDownloadEndpoint, async Task<IResult> (
```
dluc, line 307: I would make some changes to this endpoint:
* use GET method, passing all params as query. GET is much easier to test, is cacheable, and is usually how files are downloaded
* I would remove the custom headers DocID, Volume etc. Why are they needed?
* There should be Content Type and Content Length headers. The type can be fetched from memory or using MimeTypesDetection class.
* There's some confusion between "document" and "file". The new feature allows to download one file, so I would avoid using the "document" terminology, e.g. `ExportDocumentAsync` should be `ExportFileAsync`
* About status codes, why turn `KernelMemoryException` into a 400 rather than a 503?

service/Abstractions/ContentStorage/IContentStorage.cs
```diff
@@ -90,4 +90,20 @@ public interface IContentStorage
...
+    /// <summary>
+    /// Fetch file info from storage
+    /// </summary>
+    /// <param name="index">Index name</param>
+    /// <param name="documentId">Document ID</param>
+    /// <param name="fileName"></param>
+    /// <param name="logErrIfNotFound">Whether to log an error if the file does not exist. An exception will be raised anyway.</param>
+    /// <param name="cancellationToken">Async task cancellation token</param>
+    /// <returns>File content</returns>
+    Task<StreamableContentFile> GetFileInfoAsync(
```
dluc, line 103: the method is not fetching just the info but also the content, so I think we should consider reusing the existing `ReadFileAsync` method, e.g. allowing streaming and including the data needed. Or, we can limit `GetFileInfoAsync` to return just info and not the content, changing `ReadFileAsync` to return a streamable file content (without anything else).

clients/dotnet/WebClient/MemoryWebClient.cs
```diff
@@ -233,6 +235,43 @@ public async Task DeleteDocumentAsync(string documentId, string? index = null, C
...
+    {
+        DocumentQuery request = new()
+        {
+            Index = index,
+            FileName = fileName,
+            DocumentId = documentId
+        };
+        using StringContent content = new(JsonSerializer.Serialize(request), Encoding.UTF8, "application/json");
+        // TODO: use Constants.HttpDownloadEndpoint
+        HttpResponseMessage? response = await this._client.PostAsync("/download", content, cancellationToken).ConfigureAwait(false);
```
dluc, line 253: 'GET' method

clients/dotnet/WebClient/DocumentQuery.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Collections.Generic;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.KernelMemory;
+
+public class DocumentQuery
```
dluc, line 8: I think this class can be deleted

service/Abstractions/Models/StreamableFileContent.cs
```diff
@@ -0,0 +1,53 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Microsoft.KernelMemory;
+
+public sealed class StreamableFileContent : IDisposable
```
dluc, line 9: wondering if we can reuse .NET `FileInfo` class and delete this one

service/tests/Core.FunctionalTests/ServerLess/SubDirFilesAndStreamsTest.cs
```diff
@@ -78,4 +78,23 @@ public async Task ItImportsStreams()
...
+
+    [Fact]
+    [Trait("Category", "Serverless")]
+    public async Task ItAllowsToDownloadFilesWithTheSameName()
+    {
+        var doc = new Document { Id = "clones" };
+        doc.AddFile(Path.Join(this._fixturesPath, "Doc1.txt"));
+        doc.AddFile(Path.Join(this._fixturesPath, "Documents", "Doc1.txt"));
+
+        // Act - Assert no exception occurs
```
dluc, line 90:
```suggestion
        // Act
```

dluc (approved):

PR #423, opened 2024-04-22
Rename WithContentDecoder and AddContentDecoder methods

marcominerva:
`WithContentDecoder` and `AddContentDecoder` extension methods have been renamed, respectively, to `WithCustomContentDecoder` and `AddCustomContentDecoder` to better align with library naming conventions.

Existing methods have been marked as Obsolete to avoid breaking changes.


dluc:
The "WithCustom" prefix is meant for replacing default implementations, e.g. "rather than using the default X using a custom X". Maybe the rule hasn't been followed all the times, but I'd rather reduce the instances where the longer "WithCustom" is used. Maybe in future we should remove all of them, but I'd rather not spend time on these minor details.

In the case of decoders, it's both "add a decoder for an unknown type" or "rather than using default decoder X for type Y, use another X". If we want to go down this path, then we should have both, with different behavior:

* WithContentDecoder: add a decoder for a new type, previously not supported
* WithCustomContentDecoder: replace an existing decoder for type X

Considering also that decoders can decode multiple types, replacing a decoder would require comparing the supported types, and that's too much work. 

Overall, I would rather leave things as they are. If you like, we can collect a list of similar improvements, which are breaking changes by the way, and tackle all of them at once in future.

marcominerva:
OK, I get the point. I agree with you, we can take note of such improvements and then work on them when everything is consolidated.

PR #424, opened 2024-04-22
Add IKernelMemoryBuilder generic oveloads

marcominerva:
This PR adds generic overloads to the following `IKernelMemoryBuilder` extension methods, allowing to fully leverage dependency injection:

- `WithCustomIngestionQueueClientFactory`
- `WithCustomStorage`
- `WithCustomMimeTypeDetection`
- `WithCustomTextGenerator`
- `WithCustomImageOcr`
- `WithCustomPromptProvider`
- `WithCustomSearchClient`

From this list, `WithCustomEmbeddingGenerator` and `WithCustomMemoryDb` methods are missing, because providing a generic implementation of them requires a little refactoring, so it will be added later.

### DIFF
--- a/service/Abstractions/KernelMemoryBuilderExtensions.cs
+++ b/service/Abstractions/KernelMemoryBuilderExtensions.cs
@@ -79,6 +79,14 @@ public static partial class KernelMemoryBuilderExtensions
         return builder;
     }
 
+    public static IKernelMemoryBuilder WithCustomIngestionQueueClientFactory<T>(
+        this IKernelMemoryBuilder builder, T service) where T : QueueClientFactory
+    {
+        service = service ?? throw new ConfigurationException("The ingestion queue client factory instance is NULL");
+        builder.AddSingleton<QueueClientFactory, T>();
+        return builder;
+    }
+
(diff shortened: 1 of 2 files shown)


JonathanVelkeneers:
@marcominerva @dluc 
Hi, I think a mistake has been made in these changes!
If these are generic overloads to the DI methods, why is it still required to pass an instance of the type?
Here the instances are a parameter: [KernelMemoryBuilderExtensions.cs](https://github.com/microsoft/kernel-memory/pull/424/files#diff-05649cd49e9a1549e1c629e9a73bf13d076c624077ac371c7b557135695dc715)
These ones do it correctly without instances: [DependencyInjection.cs](https://github.com/microsoft/kernel-memory/pull/424/files#diff-1a503695230f35db3769b85df9f38224b73787df62705bf802b9234a0c0b5ec1)

The instances passed into KernelMemoryBuilderExtensions.cs are only checked for null and not actually used

marcominerva:
My bad... there are typos caused by copy/paste. I have opened a PR to fix it: https://github.com/microsoft/kernel-memory/pull/433.

PR #425, opened 2024-04-22
Azure queues config

marcominerva:
Currently, Azure Queues settings like message processing retries are hard-coded.

This PR moves the settings to the `AzureQueuesConfig` class.


extensions/AzureQueues/AzureQueuesConfig.cs
```diff
@@ -35,6 +35,31 @@ public enum AuthTypes
...
+    /// <summary>
+    /// How many times to dequeue a messages and process before moving it to a poison queue.
+    /// </summary>
+    public int MaxRetriesBeforePoisonQueue { get; set; } = 20;
+
+    /// <summary>
+    /// Suffix used for the poison queues.
+    /// </summary>
+    public string PoisonQueueSuffix { get; set; } = "-poison";
+
```
dluc, line 62: we should add a Validate method and call it from AzureQueuePipeline ctor, making sure that
* PollDelayMsecs > 0
* FetchBatchSize > 0
* FetchLockSeconds >= 30
* MaxRetriesBeforePoisonQueue >= 0
* PoisonQueueSuffix is not empty and valid

dluc (approved):

Issue #159, opened 2023-11-16
Azure Cognitive Search "Hybrid Search + Semantic" Search - Is it supported?


slm-2015:
Hi,

Can I use Azure Search "Hybrid Search + Semantic" Search using Kernel Memory? Or is only vector search possible?

Just looking at this article "Hybrid Search + Semantic" => (https://techcommunity.microsoft.com/t5/ai-azure-ai-services-blog/azure-cognitive-search-outperforming-vector-search-with-hybrid/ba-p/3929167) , it suggests that using this would provide better relevance in the returned memories.

Any pointers welcome.

Thanks,
Simon.

dluc:
Hi Simon, the feature is not available yet, but definitely on our radar, high priority I would say.

dluc:
Quick update: the `IMemoryDb`interface used by all memory connectors now allows to see the user input text query, so it's possible to write a custom connector using hybrid search and custom features.

crickman:
Couldn't `AzureAISearchMemory` be updated to support hybrid search?  The appsettings could be updated to control the mode (hybrid or simple vector): `AzureAISearchConfig`

luismanez:
Just chiming in here to emphasize how important is to support hybrid (and semantic) search in the library, as that would improve the RAG process a lot... 😄

dluc:
@luismanez that looks good to me, nothing against :-) I was checking Azure AI Search config to see how complex indexes can be, and there are really a lot of permutations. 

I'm assuming this change would not require changed to the index schema? KM creates indexes automatically, so I'm not sure if you're proposal should include changes to the index creation and how complex that could be.

luismanez:
OK, great!, thanks. I'll do the PR as soon as possible.

This does not require changes to the index schema. We are just enabling Hybrid search, and as far as I understand from here: https://learn.microsoft.com/en-us/azure/search/hybrid-search-how-to-query#hybrid-query-request we just need to provide the vector and search parameters to run the Hybrid query (currently only the vector is provided).

In the (short? 😄 ) future, would be great to allow also Semantic search, but that would require changes in the Index schema, and would be great to have the ability to provide a custom Schema...

dluc:
> OK, great!, thanks. I'll do the PR as soon as possible.
> 
[...]

PR #428, opened 2024-04-23
Add Azure AI Search hybrid search support

luismanez:
Already described in issue #159 
The main idea is to support Azure AI search Hybrid search.

Already described in issue #159

### DIFF
--- a/KernelMemory.sln
+++ b/KernelMemory.sln
@@ -258,6 +258,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "modules", "modules", "{C2D3
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Service.AspNetCore", "service\Service.AspNetCore\Service.AspNetCore.csproj", "{A46B0BE1-03F2-4520-A3DA-FD845BA1FD69}"
 EndProject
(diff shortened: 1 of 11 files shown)


dluc:
Made a few changes:
* fixed a bug: when using hybrid search, the connector was not stopping at `MaxMatchesCount`. Fixed the client and added a protection in SearchClient, to avoid issues with tokens and long running prompts with local models
* added an example showing the different result with hybrid search

@luismanez from a RAG perspective I don't see major improvements with hybdrid search, only a different ranking of results that might or might not affect results. The example I added works because it uses an exact string match, but it would be nice having an example where the "question" is a real question, and thanks to hybrid search the Ask method is able to produce an answer. If you have such example, it would be great adding it.

dluc (approved):

luismanez:
> @luismanez from a RAG perspective I don't see major improvements with hybdrid search, only a different ranking of results that might or might not affect results. The example I added works because it uses an exact string match, but it would be nice having an example where the "question" is a real question, and thanks to hybrid search the Ask method is able to produce an answer. If you have such example, it would be great adding it.

Hi @dluc, thanks for your changes and merging this. You did a good point here. I do not have yet a real-world sample. I'm pretty sure Hybrid helps when you ask something about a specific Product id, Reference, etc. A different ranking can make a difference when you want to save some tokens, (i.e: only the first 3 results are passed to the Model). 
Also, I'm a truly believer of this table (from here: https://techcommunity.microsoft.com/t5/ai-azure-ai-services-blog/azure-ai-search-outperforming-vector-search-with-hybrid/ba-p/3929167):

image

We should work on making Semantic search support in KM but in the meantime, at least Hybrid is supported 😄 

Thanks again!

PR #438, opened 2024-04-26
Azure AI Search: Add support for batching save operations

marcominerva:

Add support for batch support in Azure AI Search and SaveRecordsHandler.

* New `KernelMemory:DataIngestion:MemoryDbUpsertBatchSize` setting (default: 1) to manage memory records batch size.
* Implement the new optional `IMemoryDbBatchUpsert` interface in `AzureAISearchMemory`.
* Fix old bug in `AzureAISearchMemory` occurring when `EmbeddingGenerationEnabled` is `false`.

### DIFF
--- a/extensions/AzureAISearch/AzureAISearch/AzureAISearchMemory.cs
+++ b/extensions/AzureAISearch/AzureAISearch/AzureAISearchMemory.cs
@@ -3,7 +3,6 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Text.RegularExpressions;
@@ -100,6 +99,8 @@ public class AzureAISearchMemory : IMemoryDb, IMemoryDbBatchUpsert
     /// <inheritdoc />
     public Task CreateIndexAsync(string index, int vectorSize, CancellationToken cancellationToken = default)
     {
+        // Vectors cannot be less than 2 - TODO: use different index schema
+        vectorSize = Math.Max(2, vectorSize);
         return this.CreateIndexAsync(index, AzureAISearchMemoryRecord.GetSchema(vectorSize), cancellationToken);
     }
 
--- a/extensions/AzureAISearch/AzureAISearch/AzureAISearchMemory.cs
+++ b/extensions/AzureAISearch/AzureAISearch/AzureAISearchMemory.cs
@@ -30,8 +30,7 @@ namespace Microsoft.KernelMemory.MemoryDb.AzureAISearch;
 /// * support custom schema
 /// * support custom Azure AI Search logic
 /// </summary>
-[Experimental("KMEXP03")]
-public sealed class AzureAISearchMemory : IMemoryDb
+public class AzureAISearchMemory : IMemoryDb, IMemoryDbBatchUpsert
 {
     private readonly ITextEmbeddingGenerator _embeddingGenerator;
     private readonly ILogger<AzureAISearchMemory> _log;
@@ -127,13 +126,24 @@ public sealed class AzureAISearchMemory : IMemoryDb
     /// <inheritdoc />
     public async Task<string> UpsertAsync(string index, MemoryRecord record, CancellationToken cancellationToken = default)
     {
+        var result = this.BatchUpsertAsync(index, new[] { record }, cancellationToken);
+        var id = await result.SingleAsync(cancellationToken).ConfigureAwait(false);
+        return id;
+    }
+
+    /// <inheritdoc />
+    public async IAsyncEnumerable<string> BatchUpsertAsync(
+        string index,
+        IEnumerable<MemoryRecord> records,
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    {
         var client = this.GetSearchClient(index);
-        AzureAISearchMemoryRecord localRecord = AzureAISearchMemoryRecord.FromMemoryRecord(record);
+        var localRecords = records.Select(AzureAISearchMemoryRecord.FromMemoryRecord);
 
         try
         {
             await client.IndexDocumentsAsync(
-                IndexDocumentsBatch.Upload(new[] { localRecord }),
+                IndexDocumentsBatch.Upload(localRecords),
                 new IndexDocumentsOptions { ThrowOnAnyError = true },
                 cancellationToken: cancellationToken).ConfigureAwait(false);
         }
@@ -142,7 +152,10 @@ public sealed class AzureAISearchMemory : IMemoryDb
             throw new IndexNotFound(e.Message, e);
         }
 
-        return record.Id;
+        foreach (var record in records)
+        {
+            yield return record.Id;
+        }
     }
 
     /// <inheritdoc />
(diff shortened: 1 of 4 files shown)


service/Core/Handlers/SaveRecordsHandler.cs
```diff
@@ -106,63 +111,108 @@ private static string GetRecordId(string documentId, string partId)
...
-
-            foreach (IMemoryDb client in this._memoryDbs)
+            if (records.Count != 0)
             {
-                try
+                foreach (IMemoryDb client in this._memoryDbs)
                 {
-                    await this.CreateIndexOnceAsync(client, createdIndexes, pipeline.Index, record.Vector.Length, cancellationToken).ConfigureAwait(false);
+                    if (client is IBatchMemoryDb batchClient)
+                    {
```
dluc, line 163: there's a lot of code duplication here, thoughts about reducing it?

service/Core/Handlers/SaveRecordsHandler.cs
```diff
@@ -106,63 +111,108 @@ private static string GetRecordId(string documentId, string partId)
         var createdIndexes = new HashSet<string>();
 
         // For each embedding file => For each Memory DB => Upsert record
-        foreach (FileDetailsWithRecordId embeddingFile in GetListOfEmbeddingFiles(pipeline))
+        foreach (FileDetailsWithRecordId[] embeddingFiles in GetListOfEmbeddingFiles(pipeline).Chunk(this._config.DataIngestion.SaveBatchSize))
```
dluc, line 114: batching could be used also in `SavePartitionsAsync` right?

extensions/AzureAISearch/AzureAISearch/AzureAISearchMemory.cs
```diff
@@ -353,21 +353,28 @@ await foreach (SearchIndex? searchIndex in indexesAsync.ConfigureAwait(false))
         return false;
     }
 
-    private async IAsyncEnumerable<string> UpsertBatchAsync(
+    public async IAsyncEnumerable<string> UpsertBatchAsync(
```
dluc, line 356: pls move the public method up close to the other Upsert method

service/Abstractions/MemoryStorage/IBatchMemoryDb.cs
```diff
@@ -0,0 +1,25 @@
...
+
+using System.Collections.Generic;
+using System.Threading;
+
+namespace Microsoft.KernelMemory.MemoryStorage;
+
+/// <summary>
+/// Common interface MemoryDB that supports batching operations.
+/// </summary>
+public interface IBatchMemoryDb : IMemoryDb
```
dluc, line 11: I'd avoid inheritance, and keep these separate

service/Abstractions/MemoryStorage/IBatchMemoryDb.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    /// <summary>
```
dluc, line 13: pls fix the docs below, using plural etc.  Also you can remove "creates the given index if it does not exist." here and in `IMemoryDb` because that's not how we're implementing these methods, which throw IndexNotFound instead

service/Core/Handlers/SaveRecordsHandler.cs
```diff
@@ -45,6 +45,7 @@ private static string GetRecordId(string documentId, string partId)
 
     private readonly IPipelineOrchestrator _orchestrator;
     private readonly List<IMemoryDb> _memoryDbs;
+    private readonly KernelMemoryConfig _config;
```
dluc, line 48: I'd avoid exposing the entire config, using something like `private readonly int _upsertBatchSize;` instead

dluc (changes requested):
A few asks:
* We need to ship the Abstractions nuget with the new interface, before we can use it in other places. We'll need a separate PR focused only on the "Abstractions" project changes, merge it, release the nuget, etc.

* I would prefer splitting the two interfaces, not using inheritance, e.g. `class AzureAISearchMemory : IMemoryDb, IMemoryDbBatchUpsert` 

* 'UpsertAsync' can be refactored to use `BatchUpsertAsync`, avoiding duplicate code

* I would rename the new interface to `IMemoryDbBatchUpsert` and the method to `BatchUpsertAsync`

* I believe some code is missing in the handler, see comment

dluc:
I added and released the new interface, you should be able to move forward with this PR without creating another one

service/Abstractions/MemoryStorage/IBatchMemoryDb.cs
dluc: done in https://github.com/microsoft/kernel-memory/pull/441

service/Abstractions/MemoryStorage/IBatchMemoryDb.cs
dluc: done in https://github.com/microsoft/kernel-memory/pull/441

extensions/AzureAISearch/AzureAISearch/AzureAISearchMemory.cs
```diff
@@ -30,7 +30,7 @@ namespace Microsoft.KernelMemory.MemoryDb.AzureAISearch;
 /// * support custom schema
 /// * support custom Azure AI Search logic
 /// </summary>
-public class AzureAISearchMemory : IMemoryDb
+public class AzureAISearchMemory : IBatchMemoryDb
```
dluc, line 33:
```suggestion
public class AzureAISearchMemory : IMemoryDb, IMemoryDbBatchUpsert
```

extensions/AzureAISearch/AzureAISearch/AzureAISearchMemory.cs
anthonypuppo, line 356: nit: The Azure.Search.Documents package already comes with [SearchIndexingBufferedSender](/Azure/azure-sdk-for-net/blob/main/sdk/search/Azure.Search.Documents/src/Batching/SearchIndexingBufferedSender.cs).

extensions/AzureAISearch/AzureAISearch/AzureAISearchMemory.cs
```diff
@@ -144,6 +144,33 @@ public async Task<string> UpsertAsync(string index, MemoryRecord record, Cancell
         return record.Id;
     }
 
+    /// <inheritdoc />
+    public async IAsyncEnumerable<string> BatchUpsertAsync(
```
dluc, line 148: nit: refactor `UpsertAsync` (the method above) to call this one, removing code duplication/reusing the same logic

service/Core/Handlers/SaveRecordsHandler.cs
```diff
@@ -234,27 +290,21 @@ private static string GetRecordId(string documentId, string partId)
                             this._log.LogWarning(e, "Index {0} not found, attempting to create it", pipeline.Index);
                             await this.CreateIndexOnceAsync(client, createdIndexes, pipeline.Index, record.Vector.Length, cancellationToken, true).ConfigureAwait(false);
 
-                            this._log.LogTrace("Retry: saving record {0} in index '{1}'", record.Id, pipeline.Index);
+                            this._log.LogTrace("Retry: Saving record {0} in index '{1}'", record.Id, pipeline.Index);
```
dluc, line 293:
```suggestion
                            this._log.LogTrace("Retry: saving record {0} in index '{1}'", record.Id, pipeline.Index);
```

dluc:
next steps:
* rebase on latest main (I tried and there's a lot of conflicts, so I'll need more time)
* check exceptions handling
* manual tests

dluc (approved):

Issue #432, opened 2024-04-24
[Feature Request] Handle pipeline failure


marcominerva:
### Context / Scenario

Currently, pipeline failure isn't handled at all:

https://github.com/microsoft/kernel-memory/blob/775301a1cdd84ab7a54458d3fa453a8763c0744d/service/Abstractions/Pipeline/DataPipeline.cs#L434-L439

### The problem

It is important to keep track of errors that occurs during pipeline exectuion.

### Proposed solution

We can add a couple of properties to [DataPipeline.cs](https://github.com/microsoft/kernel-memory/blob/main/service/Abstractions/Pipeline/DataPipeline.cs):

```csharp
/// <summary>
/// The step that failed, if any.
/// </summary>
public string? FailedStep { get; set; } = null;

/// <summary>
/// The error that caused the pipeline to fail, if any.
/// </summary>
public string? FailureReason { get; set; } = null;
```

`FailureReason` can be useful to immediately obtain information about the problem, but it is not strictly necessary to implement this feature.

Then, we need to handle exceptions during pipeline execution, both in [InProcessPipelineOrchestrator.cs](https://github.com/microsoft/kernel-memory/blob/main/service/Core/Pipeline/InProcessPipelineOrchestrator.cs) and in [DistributedPipelineOrchestrator.cs](https://github.com/microsoft/kernel-memory/blob/main/service/Core/Pipeline/DistributedPipelineOrchestrator.cs).

Finally, after updating the [DataPipelineStatus.cs](https://github.com/microsoft/kernel-memory/blob/main/service/Abstractions/Models/DataPipelineStatus.cs) class accordingly, we just need this code:

```csharp
public DataPipelineStatus ToDataPipelineStatus()
{
    return new DataPipelineStatus
    {
        Completed = this.Complete,
        Failed = this.FailedStep != null,
        Empty = this.Files.Count == 0,
        Index = this.Index,
        DocumentId = this.DocumentId,
        Tags = this.Tags,
        Creation = this.Creation,
        LastUpdate = this.LastUpdate,
        Steps = this.Steps,
        RemainingSteps = this.RemainingSteps,
        CompletedSteps = this.CompletedSteps,
        FailedStep = this.FailedStep,
        FailureReason = this.FailureReason,
    };
}
```

### Importance

would be great to have

dluc:
What about using DataPipeline.Files.LogEntries to know what happened? work is always done on a file, and a global log can be extracted iterating the list of files.

[...]

PR #443, opened 2024-04-29
Pipeline failures

marcominerva:
See https://github.com/microsoft/kernel-memory/issues/432.

### DIFF
--- a/service/Abstractions/Models/DataPipelineStatus.cs
+++ b/service/Abstractions/Models/DataPipelineStatus.cs
@@ -67,4 +67,18 @@ public class DataPipelineStatus
     [JsonPropertyOrder(17)]
     [JsonPropertyName("completed_steps")]
     public List<string> CompletedSteps { get; set; } = new();
+
+    /// <summary>
+    /// The failed step, if any.
+    /// </summary>
+    [JsonPropertyOrder(18)]
+    [JsonPropertyName("failed_step")]
+    public string? FailedStep { get; set; } = null;
+
+    /// <summary>
+    /// The error that caused the pipeline to fail, if any.
+    /// </summary>
+    [JsonPropertyOrder(19)]
+    [JsonPropertyName("failure_reason")]
+    public string? FailureReason { get; set; } = null;
 }
--- a/service/Abstractions/Models/DataPipelineStatus.cs
+++ b/service/Abstractions/Models/DataPipelineStatus.cs
@@ -79,6 +79,6 @@ public class DataPipelineStatus
     /// The error that caused the pipeline to fail, if any.
     /// </summary>
     [JsonPropertyOrder(19)]
-    [JsonPropertyName("failure_reason")]
-    public string? FailureReason { get; set; } = null;
+    [JsonPropertyName("logs")]
+    public string? Logs { get; set; } = null;
 }
--- a/service/Abstractions/Pipeline/DataPipeline.cs
+++ b/service/Abstractions/Pipeline/DataPipeline.cs
@@ -250,11 +250,11 @@ public sealed class DataPipeline
     public List<string> CompletedSteps { get; set; } = new();
 
     /// <summary>
-    /// The step that failed, if any.
+    /// Tells if the pipeline failed.
     /// </summary>
     [JsonPropertyOrder(6)]
-    [JsonPropertyName("failed_step")]
-    public string? FailedStep { get; set; } = null;
+    [JsonPropertyName("failed")]
+    public bool Failed { get; set; } = false;
 
     /// <summary>
     /// Document tags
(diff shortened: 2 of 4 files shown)


service/Abstractions/Models/DataPipelineStatus.cs
```diff
@@ -67,4 +67,18 @@ public class DataPipelineStatus
...
+    [JsonPropertyOrder(18)]
+    [JsonPropertyName("failed_step")]
+    public string? FailedStep { get; set; } = null;
+
+    /// <summary>
+    /// The error that caused the pipeline to fail, if any.
+    /// </summary>
+    [JsonPropertyOrder(19)]
+    [JsonPropertyName("logs")]
+    public string? Logs { get; set; } = null;
```
dluc, line 83: I would avoid the "string" type which limits to a single unstructured entry, and use `List<PipelineLogEntry> LogEntries` if this is really needed. Thinking about the scenarios, I still don't see a use case though.

service/Abstractions/Pipeline/DataPipeline.cs
```diff
@@ -249,6 +249,13 @@ public string GetHandlerOutputFileName(IPipelineStepHandler handler, int index =
     [JsonPropertyName("completed_steps")]
     public List<string> CompletedSteps { get; set; } = new();
 
+    /// <summary>
+    /// Tells if the pipeline failed.
+    /// </summary>
+    [JsonPropertyOrder(6)]
+    [JsonPropertyName("failed")]
+    public bool Failed { get; set; } = false;
```
dluc, line 257: how do you ensure this is consistent? 

for instance, if ContentStorage becomes unavailable and the pipeline job expires, the value will remain `false` and never change to `true` because there's nothing left running to retry.

service/Abstractions/Pipeline/DataPipeline.cs
```diff
@@ -268,6 +275,13 @@ public string GetHandlerOutputFileName(IPipelineStepHandler handler, int index =
     [JsonPropertyName("files")]
     public List<FileDetails> Files { get; set; } = new();
 
+    /// <summary>
+    /// The Erorr logs that caused the pipeline to fail, if any.
+    /// </summary>
+    [JsonPropertyOrder(19)]
+    [JsonPropertyName("logs")]
+    public string? Logs { get; set; } = null;
```
dluc, line 283: ditto about consistency. I think a better approach would be assigning an approximate completion time and using that to establish if the pipeline failed.

service/Core/Pipeline/DistributedPipelineOrchestrator.cs
```diff
@@ -220,21 +220,48 @@ public override async Task RunPipelineAsync(DataPipeline pipeline, CancellationT
...
+            else
+            {
+                pipeline.Failed = true;
+                pipeline.Logs = string.Join(", ", pipeline.Files.Where(f => f.LogEntries is not null).SelectMany(f => f.LogEntries!).Select(l => $"{l.Source}: {l.Text}"));
+
+                this.Log.LogError("Handler {0} failed to process pipeline {1}", currentStepName, pipeline.DocumentId);
+            }
         }
-        else
+        catch (Exception ex)
```
dluc, line 251: which code is throwing this exception? why stop retrying?

service/Core/Pipeline/InProcessPipelineOrchestrator.cs
```diff
@@ -165,25 +165,57 @@ public override async Task RunPipelineAsync(DataPipeline pipeline, CancellationT
...
+                    pipeline.Failed = true;
+                    pipeline.Logs = string.Join(", ", pipeline.Files.Where(f => f.LogEntries is not null).SelectMany(f => f.LogEntries!).Select(l => $"{l.Source}: {l.Text}"));
+                    await this.UpdatePipelineStatusAsync(pipeline, cancellationToken).ConfigureAwait(false);
+
+                    this.Log.LogError("Handler '{0}' failed to process pipeline '{1}/{2}'", currentStepName, pipeline.Index, pipeline.DocumentId);
+                    throw new OrchestrationException($"Pipeline error, step {currentStepName} failed");
+                }
             }
-            else
+            catch (Exception ex)
```
dluc, line 203: ditto, where is the exception coming from? e.g. if `UpdatePipelineStatusAsync` is failing in the `try` block it will likely fail also in the `catch` block

marcominerva:
The `Logs` property was just an idea, but we can remove it for now, the important thing is how to understand if the pipeline actually failed. I have added the `catch` blocks in the **Distributed** and **InProcess** Orchestrators to handle all unexpected errors like an attempt to decode an invalid PDF file (that will cause an Exception in the `stepHandler.InvokeAsync` method).

@dluc Your idea about assigning an approximate completion time and use it to determine the status of the pipeline is quite interesting and will allow to add the failure management without touching the Orchestrators at all, but how to determine a suitable value for the completion time? Maybe it could be a setting in configuration, that is saved in the `DataPipeline` object when building it?

dluc:
Parking this for now, since nobody has raised similar concerns. 
In this area I think users will want a stateful orchestrator that exposes each job with some nice UI, allowing to inspect, to pause/cancel, to re-run etc. and KM is not meant to develop in that direction. I would leave KM to rely on logs, and potentially develop the orchestration features separately, leveraging existing solutions.

PR #448, opened 2024-04-30
New abstractions to support file download

dluc:

Supporting abstractions for new File Download feature. See PR #415

* New version 0.40
* Breaking changes on storage interface
* New methods on orchestration interface
* New methods on memory interface


coryisakson (approved):
Looks good, thank you

PR #459, opened 2024-05-04
Add SQL server memory db

kbeaugrand:
As discussed with @dluc I suggest my connector's code for SQL Server.

Migrate code from https://github.com/kbeaugrand/SemanticKernel.Connectors.Memory.SqlServer to Kernel Memory


extensions/SQLServer/SQLServer/SqlServerMemory.cs
```diff
@@ -0,0 +1,645 @@
...
+
+        if (!(await this.DoesIndexExistsAsync(index, cancellationToken).ConfigureAwait(false)))
+        {
+            // Index does not exist
+            yield break;
+        }
+
+        Embedding embedding = await this._embeddingGenerator.GenerateEmbeddingAsync(text, cancellationToken).ConfigureAwait(false);
+
+        string queryColumns = "[id], [payload], [tags]";
```
dluc, line 288: looks like this var is not used in this method, can I remove it?

kbeaugrand: No, embedding is used as the vector search at line 353 on the same method.

dluc: sorry I mean the `queryColumns` var, it's not used, is it?

kbeaugrand: My bad, you're identifying a bug in fact. I should use this var but it's in fact hard coded and the end... please let me fix it.

dluc (approved):

PR #468, opened 2024-05-08
Fix for when RangeUsed() in worksheet returns null.

jurepurgar:
I have an excel document that has a special worksheet which is not even visible in Excel. I don't know the specifics behind this, but 
`RangeUsed()` returns null for this worksheet. This PR addresses this by skipping the worksheets for which RangeUsed() returns null.

Fix for a special kind of Excel document.


dluc (approved):

dluc:
Thanks!

Issue #388, opened 2024-03-28
[Bug] Endless waiting with custom openAI endpoint


JonathanVelkeneers:
### Context / Scenario

I am trying to use the new functionality to use a custom openAI endpoint provided by #380.
I'm attempting to use a HuggingFace model that supports chat through their interference API ([example](https://huggingface.co/NousResearch/Nous-Hermes-2-Mixtral-8x7B-DPO) & [documentation](https://huggingface.co/blog/tgi-messages-api#using-inference-endpoints-with-openai-client-libraries)).

Using cUrl the result is as I expected:
**Request:**
```
curl -N https://api-inference.huggingface.co/models/NousResearch/Nous-Hermes-2-Mixtral-8x7B-DPO/v1/chat/completions \
    -X POST \
    -d '{"model":"NousResearch/Nous-Hermes-2-Mixtral-8x7B-DPO", "messages": [{"role":"user","content":"How is the weather in Antwerp, Belgium?"}]}' \
    -H "Content-Type: application/json" \
    -H "Authorization: Bearer XXXXXXXXXXXXXXXXXXXXXXXXXXXX"
```
    
**Response:**
```
{"id":"","object":"text_completion","created":1711630137,"model":"text-generation-inference/Nous-Hermes-2-Mixtral-8x7B-DPO-medusa","system_fingerprint":"1.4.3-sha-e6bb3ff","choices":[{"index":0,"message":{"role":"assistant","content":"As an AI, I do not have real-time access to weather information. However, I can give you weather forecasts and updates by using external APIs or services that provide weather data. Here's a sample forecast for Antwerp, Belgium from a popular weather API:\n\nToday:\nAntwerp, Belgium\n\nPartly cloudy with a high of 22°C (72°F) and a low of 10°C (50°"},"logprobs":null,"finish_reason":"length"}],"usage":{"prompt_tokens":20,"completion_tokens":100,"total_tokens":120}}%
```

However, when using it in Kernel-Memory it hangs indefinitely.
This is how it is configured in my Program.cs:
```
var huggingFaceConfig = new OpenAIConfig
{
    Endpoint = "https://api-inference.huggingface.co/models/NousResearch/Nous-Hermes-2-Mixtral-8x7B-DPO/v1",
    TextModel = "NousResearch/Nous-Hermes-2-Mixtral-8x7B-DPO",
    TextModelMaxTokenTotal = 4096,
[...]

PR #484, opened 2024-05-14
Workaround for HuggingFace streaming API

dluc:

HuggingFace API doesn't support top_p == 0 and doesn't stop correctly generating tokens after max_tokens.

See https://github.com/microsoft/kernel-memory/issues/388

* Add example showing how to use HF API (setting top_p to 0.01)
* Change OpenAI client to stop generating tokens after max_tokens has been reached. Tokens are counted using the provided tokenizer.

TODO: investigate whether there's a workaround that doesn't depend on the tokenizer.
TODO: consider an option for HF, to automatically manage top_p range 0. < x < 1.0


dluc:
Update: issue reported to HF project here https://github.com/huggingface/text-generation-inference/issues/1896

dluc:
Deferred, let's see if either HF or Azure can fix either the server or the client

Issue #487, opened 2024-05-14
IndexNotFound should be renamed to IndexNotFoundException


kbeaugrand:
(I just noticed that `IndexNotFound` should be called `IndexNotFoundException` - feel free to fix it, or I'll do that in a separate PR later on)

_Originally posted by @dluc in https://github.com/microsoft/kernel-memory/pull/486#discussion_r1600657132_

PR #488, opened 2024-05-14
Rename IndexNotFound to IndexNotFoundException

kbeaugrand:
Related to #487

Renames `IndexNotFound` to `IndexNotFoundException` across the codebase to align with naming conventions.


unknown-target
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft. All rights reserved.
```
dluc, line 1: something happened with the file name, it's called "unknown-target"

dluc: I think this file can be removed from the PR, right?

dluc:
The code was not building because the renaming should propagate to all the files using the exception. Fixed.

dluc (approved):

PR #503, opened 2024-05-20
Add support for Dependency Injection with ITextEmbeddingGenerator

marcominerva:
See https://github.com/microsoft/kernel-memory/issues/334. This PR replaces https://github.com/microsoft/kernel-memory/pull/444 so that it excludes unnecessary changes.

### DIFF
--- a/service/Abstractions/KernelMemoryBuilderExtensions.cs
+++ b/service/Abstractions/KernelMemoryBuilderExtensions.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
 using System;
-using Microsoft.Extensions.DependencyInjection;
 using Microsoft.KernelMemory.AI;
 using Microsoft.KernelMemory.Configuration;
 using Microsoft.KernelMemory.DataFormats;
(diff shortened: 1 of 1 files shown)


service/Abstractions/KernelMemoryBuilderExtensions.cs
```diff
@@ -137,6 +138,17 @@ public static partial class KernelMemoryBuilderExtensions
...
     }
 
+    public static IKernelMemoryBuilder WithCustomEmbeddingGenerator<T>(
+           this IKernelMemoryBuilder builder) where T : class, ITextEmbeddingGenerator
+    {
+        builder.Services.AddSingleton<ITextEmbeddingGenerator>(serviceProvider =>
+        {
+            return ActivatorUtilities.CreateInstance<T>(serviceProvider);
+        });
+
```
dluc, line 148: isn't this equivalent to `builder.Services.AddSingleton<ITextEmbeddingGenerator, T>()` ?

marcominerva: You're right, I used your original suggestion from https://github.com/microsoft/kernel-memory/issues/334, but actually we don't need the `ActivatorUtilities`.

dluc (approved):

Issue #508, opened 2024-05-23
[Bug] MemoryWebClient - Rewind stream before posting to server


westdavidr:
### Context / Scenario

I am using the Kernel Memory project's MemoryWebClient to upload a stream to the server. The stream needs to be rewound before it is posted, but currently, this step seems to be missing, causing the server to receive an empty stream.

### What happened?

When attempting to upload a stream using MemoryWebClient, the server receives an empty stream. It appears that the stream is not being rewound to its beginning before the upload, which results in the server not receiving the complete data as intended.

### Importance

a fix would make my life easier

### Platform, Language, Versions

Windows, C#, 0.51.240513.2

dluc:
thank you!

PR #509, opened 2024-05-23
Reset stream position before posting to server

westdavidr:
Bug #508 

Bug #508

### DIFF
--- a/clients/dotnet/WebClient/MemoryWebClient.cs
+++ b/clients/dotnet/WebClient/MemoryWebClient.cs
@@ -438,11 +438,18 @@ public sealed class MemoryWebClient : IKernelMemory
             // Add files to the form
             for (int i = 0; i < uploadRequest.Files.Count; i++)
             {
+                using Stream fileStream = uploadRequest.Files[i].FileContent;
+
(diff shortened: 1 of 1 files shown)


clients/dotnet/WebClient/MemoryWebClient.cs
```diff
@@ -438,11 +438,18 @@ private static (string contentType, long contentLength, DateTimeOffset lastModif
             // Add files to the form
             for (int i = 0; i < uploadRequest.Files.Count; i++)
             {
+                using Stream fileStream = uploadRequest.Files[i].FileContent;
```
dluc, line 441: I would avoid introducing another IDisposable/IAsyncDisposable object to manage. 

What about this?
```csharp
if (uploadRequest.Files[i].FileContent is { CanSeek: true, Position: > 0 })
{
    uploadRequest.Files[i].FileContent.Seek(0, SeekOrigin.Begin);
}
```

westdavidr: 👍🏼

dluc (approved):

PR #525, opened 2024-05-28
Added support for newest Microsoft.ML.Tokenizer.

alkampfergit:
Actually, if you do not specify a specific tokenizer for azure openai you will use a default implementation of GPT3 Token. The aim of this PR is using the newest Microsoft.Ml.Tokenizer, that support also count() to calculate token count without creating the array of tokens.

I've done modification only for AzureopenAI, but if the PR is ok I'll add also to the Standard OpenAi configuration.

Actually I had to modify the config file, because with Azure we have deployment name that does not identify correctly the deployment. I've added ModelName property that allow the user to specify the model used (maybe it could be possible to determine automatically the format with an api call).

This allows to create the correct tokenizer based on model and use a faster tokenizer. (Gpt-4o has a different tokenizer).

### DIFF
--- a/extensions/AzureOpenAI/AzureOpenAITextGenerator.cs
+++ b/extensions/AzureOpenAI/AzureOpenAITextGenerator.cs
@@ -13,7 +13,7 @@ using Azure.Core.Pipeline;
 using Azure.Identity;
 using Microsoft.Extensions.Logging;
 using Microsoft.KernelMemory.AI.AzureOpenAI.Internals;
-using Microsoft.KernelMemory.AI.OpenAI;
+using Microsoft.KernelMemory.AI.OpenAI.Internals;
 using Microsoft.KernelMemory.Diagnostics;
 
 namespace Microsoft.KernelMemory.AI.AzureOpenAI;
@@ -28,11 +28,11 @@ public sealed class AzureOpenAITextGenerator : ITextGenerator
     private readonly string _deployment;
 
     public AzureOpenAITextGenerator(
-        AzureOpenAIConfig config,
-        ITextTokenizer? textTokenizer = null,
-        ILoggerFactory? loggerFactory = null,
-        HttpClient? httpClient = null)
-        : this(config, textTokenizer, loggerFactory?.CreateLogger<AzureOpenAITextGenerator>(), httpClient)
+           AzureOpenAIConfig config,
+           ITextTokenizer? textTokenizer = null,
+           ILoggerFactory? loggerFactory = null,
+           HttpClient? httpClient = null)
+           : this(config, textTokenizer, loggerFactory?.CreateLogger<AzureOpenAITextGenerator>(), httpClient)
     {
     }
 
@@ -43,16 +43,6 @@ public sealed class AzureOpenAITextGenerator : ITextGenerator
         HttpClient? httpClient = null)
     {
         this._log = log ?? DefaultLogger<AzureOpenAITextGenerator>.Instance;
-
-        if (textTokenizer == null)
-        {
-            this._log.LogWarning(
-                "Tokenizer not specified, will use {0}. The token count might be incorrect, causing unexpected errors",
-                nameof(DefaultGPTTokenizer));
-            textTokenizer = new DefaultGPTTokenizer();
-        }
-
-        this._textTokenizer = textTokenizer;
 
         if (string.IsNullOrEmpty(config.Endpoint))
         {
@@ -105,6 +95,8 @@ public sealed class AzureOpenAITextGenerator : ITextGenerator
             default:
                 throw new ConfigurationException($"Azure OpenAI: authentication type '{config.Auth:G}' is not supported");
         }
+
+        this._textTokenizer = textTokenizer ?? GptTokenizerDetector.GetTokenizer(this._isTextModel, this._deployment, this._client, this._log);
     }
 
     /// <inheritdoc/>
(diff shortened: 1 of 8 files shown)


dluc:
> Is it possible, but the tokenizer you are referring to are one class for each tokenizer. The idea behind my PR is that we can have a single class that get constructed with the model like new ("gpt-4"), new ("text-embedding-ada-002") so you can avoid choosing C# type for the model, but instead you can simply use the very same c# class choosing the model on costructor.
> 
> I've create tiktoken specification file for cohere api, so it would probably be the only tokenizer class you use in the entire km :).

In the current code, the developer injects the tokenizer when using the builder, so having one class per model makes it easier, allowing to inject just the type if needed.
if we have a single class, how do we detect the model name? as you mentioned, Azure OpenAI works by deployment model, making this more complicated, but even with OpenAI, and particularly with refined models, autodetecting seems too hard, with code to maintain over time.
I think a configuration approach might be easier, where the developer chooses the tokenizer (either by name or by class). Thoughts?

dluc:
by the way, I'd like to clean up the tokenizers logic and the list of tokenizers, having a consistent approach. One challenge to address is that KM can be used with other models, e.g. LLama, Phi, Anthropic, etc so we need a design that can be adapted to different models, and different tokenizers

alkampfergit:
This is the reason, IMHO, why having only one class that can simply choose the model by a string can be the solution. When you use a LLM or an embedding model there is only one tokenization model you can use, so we need to be able to automatically diagnose the type of model then choose the right tokenizer to be 100% accurate.

For azure chat model you can call a completion api, and in the response you got the model you are using, so it is perfectly possible to automatically determine the type of the model. Then having the model as string, you can easily create an instance of the tokenizer given the name.

Screenshot 2024-05-29 at 09 27 12

For cohere models, I've managed to convert the original format to a file that can be opened by microsoft.ml.tokenizer.tiktoken class https://github.com/alkampfergit/ai-playground/blob/develop/src/python/langchainVarious/Tokenization/dotnetcohere.dib, so again we can still use the very same class for cohere models.

Since almost any LLM use the BPE algorithm or a variation of it, we only need to create the vocabulary file in the right format to be loaded by Microsoft Tokenizer and the game is done :).

I'll be OOF for 2 days, I hope to have small time in the weekend to extend the pull request with model autodetection (at least for openai and azure openai) so we can reeturn on this PR discussion next week.

alkampfergit:
Hi @dluc 

I've radically changed the pull request, actually with this new version you pay the price of a two simple call to embedding and chat completion but this will allow you to get EXACT model and then you can use the new and fastest tokenizer from Microsoft.ML namespace.

I've left the configuration that allows for the user to specify the tokenizer but I really think that it should be deprecated.

extensions/AzureOpenAI/AzureOpenAITextGenerator.cs
```diff
@@ -105,6 +95,8 @@ public sealed class AzureOpenAITextGenerator : ITextGenerator
             default:
                 throw new ConfigurationException($"Azure OpenAI: authentication type '{config.Auth:G}' is not supported");
         }
+
+        this._textTokenizer = textTokenizer ?? GptTokenizerDetector.GetTokenizer(this._isTextModel, this._deployment, this._client, this._log);
```
dluc, line 99: this would cause I/O in the ctor, causing fatal errors and service crash in case of a flaky network. It will also increase cost of running the app, with additional requests and tokens spent even when the tokenizer is not used. If the class is not a singleton, or if the code is deployed as a lambda, this would cause a request every time the code is instantiated, with extra costs.

Detecting the model this way is a nice idea, however the feature should be optional and opt-in, and should be probably outside the scope of the class, to maximize the chance of caching the value and reducing cost.

dluc (commented):
I think this deserves a design conversation before moving forward, because we have already 2 set of tokenizers and the caching logic can lead to new issues and additional costs.

extensions/AzureOpenAI/AzureOpenAITextGenerator.cs
alkampfergit: I agree, I'll move to Lazy component so the call would be done only the first time the service is really used.

alkampfergit:
> I think this deserves a design conversation before moving forward, because we have already 2 set of tokenizers and the caching logic can lead to new issues and additional costs.

Agree we can open a discussion on this subject :) I also think that this is. a core part of the library that needs a good design.

dluc:
I've refactored the code a bit, removing the old tokenizer. See https://github.com/microsoft/kernel-memory/pull/696

I think the feature here, if you want to pursue it, is about autodetecting which tokenizer to use. 
The feature should consider refined models, and it should be designed to minimize cost and errors.

Rather than the application runtime, I would extend "dotnet run setup" to autodetect the model and storing the information in the config file. E.g. extending OpenAI and Azure OpenAI config classes to include the tokenizer details.
Autodetection should be optional, allowing the user to choose manually.

When using OpenAI and default models, autodetection doesn't need an API call, because the list of models is well known.

When using OpenAI/Azure OpenAI refined models, I don't know if it is possible to detect the original model name.

alkampfergit:
> 
> I think the feature here, if you want to pursue it, is about autodetecting which tokenizer to use. The feature should consider refined models, and it should be designed to minimize cost and errors.

Agree.

> 
> Rather than the application runtime, I would extend "dotnet run setup" to autodetect the model and storing the information in the config file. E.g. extending OpenAI and Azure OpenAI config classes to include the tokenizer details. Autodetection should be optional, allowing the user to choose manually.

It could be an interesting feature.

> 
> When using OpenAI and default models, autodetection doesn't need an API call, because the list of models is well known.
> 

Agree!

> When using OpenAI/Azure OpenAI refined models, I don't know if it is possible to detect the original model name.

I've done a quick check and without a call it seems not to be possible, but I'll do a check.

dluc:
Closing for now, we can reopen it or develop the feature separately.

PR #526, opened 2024-05-28
Added bulk embedding support for azure.

alkampfergit:
Embedding in azure openai (Also standard openai) supports bulk embedding where we can reduce the number of calls requesting embedding for more than one single piece of text.

This is a really needed feature when we index lots of data (extraction from big database).

Actually this is a really simple pr where I've simple ported one class I'm using right now. 
I've added an IBulkTextEmbeddingGenerator that actually is implemented only by Azure openai embedding.
I've created an implementation of IBulkTextEmbeddingGenerator that does not rely on Semantic Kernel (Why waiting for SK to implement new feature?) but directly uses OpenAi SDK (the resulting class is simpler).

We can close the PR as is to introduce the new interface and the new class if everything seems ok (fell free to call for modification). Once the PR is closed I'll start 

1. changing injestion pipeline to use the new bulk interface
2. Create a plain openai implementation for bulk embedding


Directory.Packages.props
```diff
@@ -7,6 +7,7 @@
     <PackageVersion Include="Azure.AI.FormRecognizer" Version="4.1.0" />
     <PackageVersion Include="Azure.Extensions.AspNetCore.Configuration.Secrets" Version="1.3.1" />
     <PackageVersion Include="Azure.Identity" Version="1.11.3" />
+    <PackageVersion Include="Azure.AI.OpenAI" Version="1.0.0-beta.17" />
```
dluc, line 10: do you know when this package should go GA?

alkampfergit: Absolutely no idea, but actually Kernel Memory depends on Semantic Kernel that uses the same package. It is in beta since forever and I really does not know why :(.

Alternatively there is a library maintained by @marcominerva , that is active on Kernel Memory too that fully supports all API. That library is actively maintained by Marco that uses it in production and can be a viable alternative.

extensions/AzureOpenAI/AzureOpenAIBulkTextEmbeddingGenerator.cs
```diff
@@ -0,0 +1,103 @@
...
+using Azure.AI.OpenAI;
+using Azure.Identity;
+using Microsoft.Extensions.Logging;
+using Microsoft.KernelMemory.AI.OpenAI;
+using Microsoft.KernelMemory.Diagnostics;
+
+namespace Microsoft.KernelMemory.AI.AzureOpenAI;
+
+[Experimental("KMEXP01")]
+public sealed class AzureOpenAIBulkTextEmbeddingGenerator : IBulkTextEmbeddingGenerator
```
dluc, line 18: I would add this interface to the existing `AzureOpenAITextEmbeddingGenerator` class:

```csharp
public sealed class AzureOpenAITextEmbeddingGenerator : ITextEmbeddingGenerator, IBulkTextEmbeddingGenerator
```

so we can avoid duplicating the rest of the class. Then at runtime we can check is `IBulkTextEmbeddingGenerator` is available.

Directory.Packages.props
dluc: then perhaps let's link only the SK library, and this will come via SK, using the same version used by SK?

alkampfergit: It could be a solution, I'll check if it is feasible.

service/Abstractions/AI/IBulkTextEmbeddingGenerator.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.KernelMemory.AI;
+
+/// <summary>
+/// Some (if not all) provider for embeddings can transform more than
+/// one text chunk in a single call. This is used to limit the number
+/// of API call making the process faster.
+/// </summary>
+public interface IBulkTextEmbeddingGenerator : ITextEmbeddingGenerator
```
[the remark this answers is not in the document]
alkampfergit: Perfect!

extensions/AzureOpenAI/AzureOpenAIBulkTextEmbeddingGenerator.cs
alkampfergit: Ok sounds good, I'll be OOF for two days, then I'll implement all suggestions

alkampfergit:
@dluc  Actually I've revieved all the PR and honestly I've simplified a lot :). First of all Kernel Memory wrapper already implemented batch feature, so I can simply use that class instead of using beta version of the library. I've changed the name of the interface and added that functionality on the standard classes that uses openai.

Once this is closed I'll proceed to use the batch feature in the ingestion pipeline.

dluc (approved):

PR #531, opened 2024-06-04
Added support for batch embedding generation.

alkampfergit:
After the introduction of ITextEmbeddingBatchGenerator we need to use that interface to support batch generation of embedding in pipeline.

I'm not 100% sure of the design, but it is a think that is needed in real project in my opinion. I've introduced a EmbeddingGeneratorHelper that can handle generation of embeddings given a list of embedding generators and block of text. Then once you retrieved the embedding a simple interface allows it to retrieve from an in memory dictionary.

Actually this object can be expanded so it can persist embeddings in a durable medium. (this will be a future extension).  This because in a real production project we often have documents that changes during the lifetime, and sometimes we will reindex big amount of text where the vast amount of text partitions does not check (not to take into consideration duplicate documents). Being able to persist embedding given a piece of text and an embedding generator is nice because it can greatly reduce the number of external call.

This is the reason why I externalized embedding generation to a special class. 

[Edit]: I've added a setting where we can specify the maximum number of element in a single batch call (openai on azure with ada limits to 16 elements).

Added also some basic tests to verify the behavior, introduced Moq as Mock library.


extensions/AzureOpenAI/AzureOpenAIConfig.cs
```diff
@@ -74,6 +74,15 @@ public enum APITypes
     /// </summary>
     public int? EmbeddingDimensions { get; set; }
 
+    /// <summary>
+    /// Old AI model like ada (https://learn.microsoft.com/en-us/azure/ai-services/openai/reference) does
+    /// not support in azure version more than 16 element in a single batch, while openai accepts up to
+    /// 2k. This settings is needed to avoid sending too large batches to the service and having an error.
+    ///
+    /// Openai Has much larger windows as explained here https://platform.openai.com/docs/api-reference/embeddings/create.
```
dluc, line 82: Could you clarify? The size of text is a different limit than the size of the batch.
OpenAI text-embedding-ada-002 accepts in input strings 8192 tokens.

extensions/AzureOpenAI/AzureOpenAIConfig.cs
```diff
@@ -74,6 +74,15 @@ public enum APITypes
...
+    /// </summary>
+    public int EmbeddingBatchMaxSize { get; set; } = 16;
```
dluc, line 84: A safe default value should be "1". Depending on which model, the region and when the model was deployed, some deployments support only a batch of 1:

> Too many inputs. The max number of inputs is 1.  We hope to increase the number of inputs per request soon. Please contact us through an Azure support request at: https://go.microsoft.com/fwlink/?linkid=2213926 for further questions.

By the way, with one deployment I was able to send batches of 2000+ strings, this is going to be very hard to manage...

service/tests/Core.FunctionalTests/Core.FunctionalTests.csproj
```diff
@@ -23,6 +23,7 @@
         </PackageReference>
         <PackageReference Include="Microsoft.Extensions.DependencyInjection" />
         <PackageReference Include="Microsoft.NET.Test.Sdk" />
+        <PackageReference Include="Moq" />
```
dluc, line 26: functional tests should not mock dependencies. If you added some tests with mocks, they should be moved under unit tests

dluc (changes requested):
Some comments to address. 

I would also avoid static classes like the new "helper", which seems like to exist to allow testing internals. For testing you can enable "internal" access, although I would avoid coupling tests with internal behavior that can be refactored without affecting public method, because tests would fail even when there's no bugs

extensions/AzureOpenAI/AzureOpenAIConfig.cs
alkampfergit: From documentation, Ada in azure accepts a batch of Max 16 elements, regardless the number of tokens

extensions/AzureOpenAI/AzureOpenAIConfig.cs
alkampfergit: I know, older deployment with the first model does not support batching, newest embedding have no limit, nevertheless the user should be able to set this value accordingly to actual deployment, I does not see a different solution

service/tests/Core.FunctionalTests/Core.FunctionalTests.csproj
alkampfergit: You are right, I'll move on the right project

alkampfergit:
> Some comments to address.
> 
> I would also avoid static classes like the new "helper", which seems like to exist to allow testing internals. For testing you can enable "internal" access, although I would avoid coupling tests with internal behavior that can be refactored without affecting public method, because tests would fail even when there's no bugs

The helper class is used because we have two class that actually create embedding from segments, to avoid code duplication I've moved code to generate embeddings (batch and non batch depending on the capabilities of registered embeddings) in that helper class.

Tests were written mainly to verify that the helper class is behaving as expected :), I know that it is an internal class, but if we need to refactor that class having a couple of test can help avoiding regressions.

If you think that those test are not necessary we can remove without any problem.

service/Core/Handlers/GenerateEmbeddingsHandler.cs
```diff
@@ -74,6 +74,12 @@ public sealed class GenerateEmbeddingsHandler : IPipelineStepHandler
         this._log.LogDebug("Generating embeddings, pipeline '{0}/{1}'", pipeline.Index, pipeline.DocumentId);
 
         var partitionsFound = false;
+
+        //we have two different strategies for generate embedding, because we need to support bulk embedding
+        //generation where we optimize the embedding generation greatly reducing the number of call to the model.
+        //we will calculate all embedding in a dictionary where the text is the key and the embedding is the value.
+        var embeddingCache = await EmbeddingGeneratorHelper.GetEmbeddingsAsync(pipeline, this, this._embeddingGenerators, this._orchestrator, cancellationToken).ConfigureAwait(false);
```
dluc, line 81: This approach needs to be revisited because the embedding data is lost in case of storage errors below. Not sure yet, I need to think about it before moving forward, I might merge the PR and fix it separately

dluc: I'm almost done updating the code, will push later after I finish testing

service/Core/Handlers/EmbeddingGeneratorHelper.cs
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    private static bool BatchReachedMaximumNumberOfElements(
+        ITextEmbeddingGenerator generator,
+        ITextEmbeddingBatchGenerator batchGenerator,
+        List<string> batch,
+        int tokenCount,
+        int pieceTokenCount)
+    {
+        return tokenCount + pieceTokenCount > generator.MaxTokens || batch.Count == batchGenerator.MaxBatchSize;
```
dluc, line 131: generator.MaxTokens (8191 with ada2) applies to individual strings only, not the batch size. 
For example, in one of my tests I sent a batch of 96 strings for a total of 83380 tokens.

dluc:
I rewrote the handler cleaning up the code accumulated over time, introducing a base class for the code in common between the normal embedding handler and the one using parallel processing. For later we can easily merge the two with some config option. 

I added support for `RequestContext` (via `custom_embedding_generation_batch_size_int`), so the batch size can be overridden during a request, for those scenario where the config value is too high and causing errors, or too low and one might want to try boosting it.

While I'd love having unit tests, it would have taken too long to rewrite the tests given the code is completely rewritten. If you want to resubmit them please do :-)

I made some tests and found an old bug that would occur when using multiple embedding generators, so I fixed that too. In my tests with Azure I could use a batch of 96 strings totaling to 80k tokens. The response is not immediate, but it works 👍

Thank you for all the work done and the patience to get this through!

alkampfergit:
Thanks! For unit test ... there is no problem :) I wrote mainly to verify that the code I wrote was functioning correctly without the need to actually call a real service. If I need to do some other PR on that specific code I'll rewrite some test if needed.

PR #596, opened 2024-06-04
Add AWS S3 Document Storage (WIP)

koteus:
At my job we are using our company's internal cloud storage solution which is compatible to AWS S3 API and I had to implement custom storage. With this PR I want to make it available to other users.

Another motivation is to make it available in configuration for KernelMemory service, because right now it is limited only to Azure Blobs and this is a blocker for us to use KernelMemory to asynchronously process large files.

I took a look in the code for Azure Blobs and adapted it to AWS S3. By using official nuget package AWSSDK.S3 allows to use it with AWS as long as with any compatible system (like MinIO).

Note that I marked this PR as WIP, because this is my first time contributing to Dotnet project and for sure I am missing some things like tests, configs etc, I will be glad if you help me point what is missing from my PR and I will add it.

Also feel free to suggest better names for classes, and any other comments are welcome!


koteus:
@microsoft-github-policy-service agree [company="Veepee"]

> @koteus please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree [company="Veepee"]

koteus:
@microsoft-github-policy-service agree [company="Veepee"]

koteus:
@microsoft-github-policy-service agree company="Veepee"

dluc (changes requested):
This looks great, thank you. I fixed a few minor things in the code style. There are a few things to address:

* When invoking async methods, the call should include `.ConfigureAwait(false)` to to prevent await from capturing the current synchronization context and causing deadlocks
* all async methods should support cancellation tokens, to allow shutting down the execution without delays/errors
* for those async methods that don't run any async code, `await Task.Run` is not necessary, you can use `return Task.CompletedTask` or `Task.FromResult(...)`
* when referencing local non-static vars/methods, pls remember to add the `this.` prefix, just for consistency with the project code style (I fixed a few, there's a few more to fix)
* async methods' name should have an `Async` suffix, it's a C# convention
* around line 200 there's one `using` statements that should be turned into `await using` -- however, that syntax is particularly tricky, so if you get stuck, I can fix it before merging

For all the above, you should find examples in other classes, e.g. code style used in `AzureBlobsStorage` and `PostgresDbClient` might help.

service/Service/ServiceConfiguration.cs
```diff
@@ -206,20 +210,20 @@ private void ConfigureIngestionEmbeddingGenerators(IKernelMemoryBuilder builder)
             {
                 case string x when x.Equals("AzureOpenAI", StringComparison.OrdinalIgnoreCase):
                 case string y when y.Equals("AzureOpenAIEmbedding", StringComparison.OrdinalIgnoreCase):
-                {
-                    var instance = this.GetServiceInstance<ITextEmbeddingGenerator>(builder,
-                        s => s.AddAzureOpenAIEmbeddingGeneration(this.GetServiceConfig<AzureOpenAIConfig>("AzureOpenAIEmbedding")));
-                    builder.AddIngestionEmbeddingGenerator(instance);
-                    break;
-                }
+                    {
```
koteus, line 213: @dluc Do you use some kind of formatter on this project? This spaces were added by my vscode and I don't know how to configure it not to do so. Is there any tool in C# world like Prettier is in javascript?

extensions/AWS/S3/S3Storage.cs
```diff
@@ -0,0 +1,346 @@
...
+                    BucketName = this._bucketName,
+                    Key = objName,
+                },
+                cancellationToken
+            ).ConfigureAwait(false))
+            {
+                var memoryStream = new MemoryStream();
+                var responseStream = response.ResponseStream;
+
+                await using (responseStream.ConfigureAwait(false))
```
koteus, line 186: @dluc is this the correct way to use `await using`? And should we also do something similar with `memoryStream`? (I am not sure when to dispose it, because it is used in the callback function of StreamableFileContent, line 197)

koteus:
> This looks great, thank you. I fixed a few minor things in the code style. There are a few things to address:
> 
> * When invoking async methods, the call should include `.ConfigureAwait(false)` to prevent await from capturing the current synchronization context and causing deadlocks
> * all async methods should support cancellation tokens, to allow shutting down the execution without delays/errors
> * for those async methods that don't run any async code, `await Task.Run` is not necessary, you can use `return Task.CompletedTask` or `Task.FromResult(...)`
> * when referencing local non-static vars/methods, pls remember to add the `this.` prefix, just for consistency with the project code style (I fixed a few, there's a few more to fix)
> * async methods' name should have an `Async` suffix, it's a C# convention
> * in `S3Storage.cs` around line 200 there's one `using` statement that should be turned into `await using` -- however, that syntax is particularly tricky, so if you get stuck, I can fix it before merging
> 
> For all the above, you should find examples in other classes, e.g. code style used in `AzureBlobsStorage` and `PostgresDbClient` might help.

All these issues are fixed now. Also I was able to build the whole KernelMemory locally and reference it from my project like this:
```
<!-- <PackageReference Include="Microsoft.KernelMemory.Core" Version="0.61.240519.2" /> -->
<ProjectReference Include="..\..\kernel-memory\service\Core\Core.csproj" />
```
and I checked on my localhost that it works and uploads files to S3.

What I will try next is to add config to KernelMemory service, so I could run it in Docker providing S3 and RabbitMQ configuration.

extensions/AWS/S3/S3Storage.cs
```diff
@@ -0,0 +1,346 @@
...
+            this._log.LogCritical(msg);
+            throw new DocumentStorageException(msg);
+        }
+
+        this._bucketName = config.BucketName;
+    }
+
+    public void Dispose()
+    {
+        this._client?.Dispose();
```
koteus, line 65: Not sure if it makes sense making it disposable. I also added IDisposable to the class:
``` 
public class S3Storage : IDocumentStorage, IDisposable
```

extensions/AWS/S3/S3Storage.cs
```diff
@@ -0,0 +1,346 @@
...
+
+    /// <summary>
+    /// Generates a pre-signed URL for accessing an object stored in S3 for a limited time.
+    /// </summary>
+    /// <param name="index">The index directory in which the object resides.</param>
+    /// <param name="documentId">The document directory under the index where the object resides.</param>
+    /// <param name="fileName">The name of the file/object for which to generate the URL.</param>
+    /// <param name="validDuration">The duration for which the URL should remain valid.</param>
+    /// <returns>A pre-signed URL for accessing the object.</returns>
+    public async Task<string> GeneratePreSignedURLAsync(string index, string documentId, string fileName, TimeSpan validDuration)
```
koteus, line 223: Also this method is not part of IDocumentStorage interface, I need it in my project to be able to show links to the uploaded files as citations in the LLM answers. Do you have any advice how to stay in conformation with IDocumentStorage and still have this feature?

service/Service/ServiceConfiguration.cs
dluc: Which IDE/editor do you use?

Usually I work with Rider, that integrates Resharper (a VisualStudio extension). Rider and Resharper read and apply settings from two files: `KernelMemory.sln.DotSettings` and `.editorconfig`. 

If you use VS Code with **EditorConfig** you should get most of the formatting, though some might be missing because they come from the `.DotSettings` file.

extensions/AWS/S3/S3Storage.cs
dluc: not sure if that builds, combining `await using` and `ConfigureAwait(false)` in the same line is not possible as far as I know. `using` takes care of disposing the object for you, though in some cases dispose is an async method so you need to use `await using`. In your code I saw only one instance of this. If you check the postgresdbclient you might notice I split the code in two lines. In any case, I can fix this before merging the PR (right now I don't remember the exact syntax, sorry)

extensions/AWS/S3/S3Storage.cs
dluc: you'll need it only if you have some internal disposable references. If `_client` is a disposable object, that looks good

extensions/AWS/S3/S3Storage.cs
dluc: As a rule of thumb I would avoid extra public methods in the repo, to reduce the cost of maintaining it - perhaps consider a static extension method. I'll take a look any way, there's always exceptions :-)

extensions/AWS/S3/S3Storage.cs
koteus: I was able to build it and it actually works, it save the file to S3. I found this syntax in the codebase 
image

dluc: LGTM

dluc (changes requested):
There's a couple of problems:

### 1. Path with duplication

I created a bucket named "kernelmemory" with this config
```csharp
new AWSS3Config
{
    Auth = AWSS3Config.AuthTypes.AccessKey,
    AccessKey = "....",
    SecretAccessKey = "....",
    BucketName = "kernelmemory",
    Endpoint = "https://s3.us-west-2.amazonaws.com"
}
```
and uploaded some files in the `default` index. For some reason the bucket name is used also in the path:

image

### 2. Deletion not working, throwing exception

I ran example 002 with S3 and the last step, deleting files, fails with an exception:

```log
Deleting memories derived from 04b4132592ea4d6c99b44f2faeb6b8d3202406081139323565100
Unhandled exception. Amazon.S3.AmazonS3Exception: The specified key does not exist.
 ---> Amazon.Runtime.Internal.HttpErrorResponseException: Exception of type 'Amazon.Runtime.Internal.HttpErrorResponseException' was thrown.
```

extensions/AWS/S3/S3Storage.cs
koteus: ✅ I removed this method. In my project I had to implement it separately with another instance of AmazonS3Client because if I create extension on this S3Storage it still needs to access private variables which is not allowed by compiler.

koteus:
The 

> There's a couple of problems:
> 
> ### 1. Path with duplication
> I created a bucket named "kernelmemory" with this config
> 
> ```cs
> new AWSS3Config
> {
>     Auth = AWSS3Config.AuthTypes.AccessKey,
>     AccessKey = "....",
>     SecretAccessKey = "....",
>     BucketName = "kernelmemory",
>     Endpoint = "https://s3.us-west-2.amazonaws.com"
> }
> ```
> 
> and uploaded some files in the `default` index. For some reason the bucket name is used also in the path:
> 
> image
> ### 2. Deletion not working, throwing exception
> I ran example 002 with S3 and the last step, deleting files, fails with an exception:
> 
> ```
> Deleting memories derived from 04b4132592ea4d6c99b44f2faeb6b8d3202406081139323565100
> Unhandled exception. Amazon.S3.AmazonS3Exception: The specified key does not exist.
>  ---> Amazon.Runtime.Internal.HttpErrorResponseException: Exception of type 'Amazon.Runtime.Internal.HttpErrorResponseException' was thrown.
> ```

Both of these errors are fixed now, they were related to the fact that our company's S3 cloud is configured in such a way `bucket-name.custom-domain.com/bucket-name` so you have to repeat it twice. I fixed it now and I tested on my localhost with MinIO, it can now create and delete documents, also it works with my company's S3 cloud when I changed environment variables.

extensions/AWS/S3/AWSS3Storage.cs
```diff
@@ -0,0 +1,261 @@
...
+
+            request.ContinuationToken = response.NextContinuationToken;
+        } while (true);
+
+        foreach (var obj in allObjects)
+        {
+            var fileName = obj.Key.Trim('/').Substring(prefix.Trim('/').Length).Trim('/');
+
+            // Don't delete the pipeline status file
+            if (fileName == Constants.PipelineStatusFilename) { continue; }
```
koteus, line 236: Just to be sure when deleting document from the database we delete everything from the storage except this file `__pipeline_status.json`, right? I saw this in the Azure Blob implementation. Btw, is there a reason to keep this file?

service/Service/appsettings.json
```diff
@@ -269,6 +269,22 @@
         // Setting used only for country clouds
         "EndpointSuffix": "core.windows.net"
       },
+      "S3": {
```
koteus, line 272: This needs to be named AWSS3, right? And the keys should be according to the latest rename by you @dluc 

I added this as part of configuration to enable running Kernel Memory as a service with S3 storage. Is there anything else that we need to change for that?

service/Service/ServiceConfiguration.cs
```diff
@@ -162,6 +162,10 @@ private void ConfigureStorageDependency(IKernelMemoryBuilder builder)
                                                                 ?? this.GetServiceConfig<AzureBlobsConfig>("AzureBlob"));
                 break;
 
+            case string x when x.Equals("S3", StringComparison.OrdinalIgnoreCase):
```
koteus, line 165: This probably also needs to be changed to AWSS3 ?

And also probably some settings in `tools/interactiveSetup/` can be changed to add AWSS3 support

extensions/AWS/S3/AWSS3Storage.cs
dluc: deleting the file causes an error in the pipeline orchestration, it would take some rewrite to sort out so for now we settled on leaving it in place. If leaving these files behind becomes a problem the current suggestion is to have a periodic purge process out of band, identifying and deleting these folders.

service/Service/appsettings.json
dluc: yes AWSS3 would be ideal, thanks 👍

koteus:
Hello @dluc! So I also added configuration for appsettings generation wizard. And I built the docker image and I am able to run it and use Kernel Memory as a service from my code by just swapping `MemoryServerless` with `MemoryWebClient` instance. It seems to be working fine.

I think all issues are resolved now, do you think there is anything else I could do?

examples/002-dotnet-Serverless/Program.cs
```diff
@@ -44,6 +45,8 @@ public static async Task Main()
             .BindSection("KernelMemory:Services:AzureAIDocIntel", azDocIntelConfig)
             .BindSection("KernelMemory:Services:AzureAISearch", azureAISearchConfig)
             .BindSection("KernelMemory:Services:AzureBlobs", azureBlobConfig)
+            .BindSection("KernelMemory:Services:AzureBlobs", azureBlobConfig)
```
dluc, line 48:
```suggestion
```

dluc (approved):

Issue #408, opened 2024-04-16
[Bug] Endless pipeline loop when trying to import corrupted document


marcominerva:
### Context / Scenario

In service mode (using a queue), when trying to import a corrupted document (i.e., invalid PDF file), decoder will throw an exception, but then the message will be put again in the queue, generating an endless loop:

```
warn: Microsoft.KernelMemory.Pipeline.Queue.DevTools.SimpleQueues[0]
      Message '20240416.100334.6972903.f0f10e85d780488d927f6c03891d20d5' processing failed with exception, 
      putting message back in the queue.
      Message content: {"index":"default","document_id":"98d5b5a2d38e4cf49dbda5b50c4502c2202404161003346000416","execution_id":"e6f9118e170745168f1d0bc1a9cd39b5","steps":["extract","partition","gen_embeddings","save_records"]}
```

### What happened?

If decoding fails with an exception, I expect that the document will be marked as `failed to process` and not process again. I see also that in the [DataPipelineStatus ](https://github.com/microsoft/kernel-memory/blob/main/service/Abstractions/Pipeline/DataPipeline.cs)class, there is a `Failed` property, but it is always `false`:

https://github.com/microsoft/kernel-memory/blob/babedc0ef8aeceacc61d607f2fe5fd43d27aa3f8/service/Abstractions/Pipeline/DataPipeline.cs#L432-L448

### Importance

I cannot use Kernel Memory

### Platform, Language, Versions

Kernel Memory v0.36.240415.2

### Relevant log output

```shell
dbug: Microsoft.KernelMemory.Handlers.TextExtractionHandler[0]
      Extracting text, pipeline 'default/577dbc0439a74a0b8a4996be204b45e3202404161021070293739'
dbug: Microsoft.KernelMemory.Handlers.TextExtractionHandler[0]
      Extracting text from file 'Iceberg.pdf' mime type 'application/pdf' using extractor 'Microsoft.KernelMemory.DataFormats.Pdf.PdfDecoder'
dbug: Microsoft.KernelMemory.DataFormats.Pdf.PdfDecoder[0]
      Extracting text from PDF file
warn: Microsoft.KernelMemory.Pipeline.Queue.DevTools.SimpleQueues[0]
[...]

PR #648, opened 2024-06-05
Add poison queues support for RabbitMQ

marcominerva:
See #408.


extensions/RabbitMQ/RabbitMqConfig.cs
```diff
@@ -37,4 +40,55 @@ public class RabbitMqConfig
...
+    public string PoisonQueueSuffix { get; set; } = "-poison";
+
+    /// <summary>
+    /// Verify that the current state is valid.
+    /// </summary>
+    public void Validate()
+    {
+        if (this.MaxRetriesBeforePoisonQueue < 0)
+        {
+            throw new ConfigurationException($"RabbitQM: {nameof(this.MaxRetriesBeforePoisonQueue)} cannot be a negative number");
```
dluc, line 61:
```suggestion
            throw new ConfigurationException($"RabbitMQ: {nameof(this.MaxRetriesBeforePoisonQueue)} cannot be a negative number");
```

extensions/RabbitMQ/RabbitMqConfig.cs
```diff
@@ -37,4 +40,55 @@ public class RabbitMqConfig
...
+            throw new ConfigurationException($"RabbitMQ: {nameof(this.PoisonQueueSuffix)} is empty");
+        }
+
+        if (!IsValidQueueName(this.PoisonQueueSuffix))
+        {
+            throw new ConfigurationException($"RabbitMQ: {nameof(this.PoisonQueueSuffix)} is too long or contains invalid chars");
+        }
+    }
+
+    private static bool IsValidQueueName(string name)
```
dluc, line 75: I would do these checks directly in `Validate()`, throwing two different and specific error messages -- that would make it easier to address the error, otherwise people will have to check this class to understand how

dluc (changes requested):
a couple of minor changes. I haven't tested/fully reviewed the logic yet

extensions/RabbitMQ/RabbitMqConfig.cs
```diff
@@ -37,4 +40,45 @@ public class RabbitMqConfig
...
+        if (string.IsNullOrWhiteSpace(this.PoisonQueueSuffix))
+        {
+            throw new ConfigurationException($"RabbitMQ: {nameof(this.PoisonQueueSuffix)} is empty");
+        }
+
+        // Queue names must follow the rules described at
+        // https://www.rabbitmq.com/docs/queues#names.
+        if (this.PoisonQueueSuffix.StartsWith("amq.", StringComparison.InvariantCulture))
+        {
+            throw new ConfigurationException($"RabbitMQ: {nameof(this.PoisonQueueSuffix)} cannot start with 'amp.', as it reserved for internal use");
```
dluc, line 73:
```suggestion
            throw new ConfigurationException($"RabbitMQ: {nameof(this.PoisonQueueSuffix)} cannot start with 'amp.', as it's reserved for internal use");
```

extensions/RabbitMQ/RabbitMQPipeline.cs
```diff
@@ -112,23 +126,50 @@ public void OnDequeue(Func<string, Task<bool>> processMessageAction)
...
+                    }
                 }
                 else
                 {
-                    this._log.LogWarning("Message '{0}' failed to process, putting message back in the queue", args.BasicProperties.MessageId);
-                    this._channel.BasicNack(args.DeliveryTag, multiple: false, requeue: true);
+                    this._log.LogWarning("Message '{0}' has reached the maximum number of retries, moving to poison queue", args.BasicProperties.MessageId);
+
+                    this._channel.BasicNack(args.DeliveryTag, multiple: false, requeue: false);
+                    this.PublishMessage(this._poisonQueueName, args.Body, args.BasicProperties.MessageId, expirationMsecs: null, dequeueCount);
```
dluc, line 170: I thought we were going to use quorum queues, leaving it to RabbitMQ to manage the queue transaction. 

Doing it with code is not 100% reliable:

* if `PublishMessage` runs **after** `BasicNack`
  * if `BasicNack` succeeds and `PublishMessage` fails, the message is lost, and there won't be any retry
* if `PublishMessage` runs **before** `BasicNack`
  * if `PublishMessage` succeeds and `BasicNack` fails, the message is duplicated. If the failure continues duplicate messages will build up, potentially bringing down the service

dluc: reference for dead letter queues: https://www.rabbitmq.com/docs/quorum-queues#poison-message-handling

marcominerva: Actually, I already proposed to use Quorum Queues, but you discarded the idea in https://github.com/microsoft/kernel-memory/issues/408#issuecomment-2099251014, mainly because they don't support a delayed delivery.

Giving that, I can modify this PR to use Quorum Queues instead, without the delayed retry logic.

dluc: the other comment was about the retry strategy while the PR here is about poison queues. I think there's some misunderstanding, also because of rabbitmq complexity, about what missing features 
* increasing backoff while retrying
* discard messages after N retries
* store discarded messages in poison queues

marcominerva: Yes, I agree that RabbitMQ introduces some complexities and I think that we need to better define the requirements, because in this moment I can't get how to use Quorum Queues together with increasing backoff while retrying and the your other points.

marcominerva:
I have updated this PR using Quorum Queues and Dead Letter Queues.

There are a couple of issues:

1. After creating the Quorum Queue:

```csharp
this._channel.QueueDeclare(
    queue: this._queueName,
    durable: true,
    exclusive: false,
    autoDelete: false,
    arguments: new Dictionary<string, object>
    {
        ["x-queue-type"] = "quorum",
        ["x-delivery-limit"] = this._config.MaxRetriesBeforePoisonQueue,
        ["x-dead-letter-exchange"] = poisonExchange
    });
```

The delivery limit cannot be changed, otherwise an exception will be thrown. So, we cannot modify the value of `MaxRetriesBeforePoisonQueue`. To make this value dynamic, we should use a policy, but it seems that policies aren't supported by the RabbitMQ .NET client.

2. Messages are requeued instantly, because there isn't a native way to have a delayed retry.

marcominerva:
Hi @dluc!
Have you got the chance to think about the issues I mentioned? Do you have any idea about how to handle them?

marcominerva:
Hi @dluc,

Given that it is not possible to change the **x-delivery-limit** value after queue creation, I have added a `try... catch` block so log the error and throw an exception that explains the problem:

https://github.com/marcominerva/kernel-memory/blob/c2f2c631757cca5f1ee093a40239968b6eb4b76d/extensions/RabbitMQ/RabbitMQPipeline.cs#L74-L92

Let me know if it could be a solution.

marcominerva:
@dluc, GitHub signals that this PR has 1 change requested, but it seems that all the request changes have been addressed, or am I missing something?

dluc (approved):

dluc:
@marcominerva I think the PR is ready, sorry about the long delay. About the issue with max retries, I think we can live with it for now. The only option I see would be removing "x-delivery-limit" and handling the count manually, checking the "x-delivery-count" header. It would be easy, but let's see if anyone complains about this first.
I've made a few changes:
* log the number of attempts
* log an error on the last attempt
* enable the consumer only after all queues and exchange are setup
* change exchange suffix, make it shorter to allow longer queue names
* some more validations
* some more logging
* added a test application showing messaging coming out of the dead letter queue

dluc (approved):

PR #652, opened 2024-06-06
Add streaming response (text only) as discussed in #625

chaelli:


dluc:
hi @chaelli there's already a PR in progress for streaming responses here https://github.com/microsoft/kernel-memory/pull/400

chaelli:
@dluc I know - that's why I made it "draft" - didn't know this will also list it here. I just wanted to quickly show the "simple" way. Will close this. Also added a comment on the discussion #625 
just hope we can get some streaming solution - for direct usage in UIs this is very important

PR #660, opened 2024-06-11
Fix reading StopSequences from the config

Foorcee:
## Motivation and Context
Currently, StopSequences initialized with Array.Empty<T>() are not deserialized correctly, resulting in an empty list even when values are provided in the configuration. 

## High level description
By changing the initialization to use new List<T>(), we ensure that the StopSequences are properly deserialized from the configuration. For more details, I have created a issue: [dotnet/runtime#103282](https://github.com/dotnet/runtime/issues/103282).


Foorcee:

dluc (approved):

PR #684, opened 2024-07-03
fix: postgres search query does not use index

roldengarm:
I've imported about 6 million documents with PostgresDB as backend. I've created an HNSW index. When doing a search (using WebClient.Ask/Search) I keep getting timeouts. Upon further investigation, I saw that the query used in Kernel Memory is not optimisable as it's doing an inline calculation of "1 - the difference".

After I removed the "1 - " in our fork & deployed it, it was successfully using the index and I would get a response within 5 - 10 seconds.
I've used pgAdmin4 to confirm that the new query is using the index.

See also discussion with @dluc here: https://github.com/microsoft/kernel-memory/discussions/663#discussioncomment-9952593

### DIFF
--- a/extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
+++ b/extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
@@ -404,8 +404,8 @@ internal sealed class PostgresDbClient : IDisposable
 
         // Column names
         string columns = withEmbeddings ? this._columnsListWithEmbeddings : this._columnsListNoEmbeddings;
-        string colDifference = "difference";
-        string colMaxDifference = "@__max_difference";
+        string colDistance = "distance";
+        string colMaxDistance = "@__max_distance";
 
         // Filtering logic, including filter by similarity
         filterSql = filterSql?.Trim().Replace(PostgresSchema.PlaceholdersTags, this._colTags, StringComparison.Ordinal);
@@ -413,12 +413,12 @@ internal sealed class PostgresDbClient : IDisposable
         {
             filterSql = "TRUE";
         }
-        var maxDifference = 1 - minSimilarity;
-        filterSql += $" AND {this._colEmbedding} <=> @embedding < {maxDifference}";
+        var maxDistance = 1 - minSimilarity;
+        filterSql += $" AND {this._colEmbedding} <=> @embedding < {maxDistance}";
 
         if (sqlUserValues == null) { sqlUserValues = new(); }
 
-        sqlUserValues[colMaxDifference] = minSimilarity;
+        sqlUserValues[colMaxDistance] = minSimilarity;
 
         this._log.LogTrace("Searching by similarity. Table: {0}. Threshold: {1}. Limit: {2}. Offset: {3}. Using SQL filter: {4}",
             tableName, minSimilarity, limit, offset, filterSql);
@@ -437,10 +437,10 @@ internal sealed class PostgresDbClient : IDisposable
                 // Furthermore, colDifference can't be used in the WHERE clause
                 // as that causes a "table cannot be found error"
                 cmd.CommandText = @$"
-                SELECT {columns}, {this._colEmbedding} <=> @embedding AS {colDifference}
+                SELECT {columns}, {this._colEmbedding} <=> @embedding AS {colDistance}
                 FROM {tableName}
                 WHERE {filterSql}
-                ORDER BY {colDifference} ASC
+                ORDER BY {colDistance} ASC
                 LIMIT @limit
                 OFFSET @offset
             ";
@@ -464,8 +464,8 @@ internal sealed class PostgresDbClient : IDisposable
                     {
                         while (await dataReader.ReadAsync(cancellationToken).ConfigureAwait(false))
                         {
-                            double difference = dataReader.GetDouble(dataReader.GetOrdinal(colDifference));
-                            double similarity = 1 - difference;
+                            double distance = dataReader.GetDouble(dataReader.GetOrdinal(colDistance));
+                            double similarity = 1 - distance;
                             result.Add((this.ReadEntry(dataReader, withEmbeddings), similarity));
                         }
                     }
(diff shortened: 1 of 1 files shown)


marcominerva:
I think that, as this PR changes the returned `similarityActualValue` from PostgresSQL, also the `minSimilarity `usage must be adjusted.

In fact, going deep in the code, I see this:

https://github.com/microsoft/kernel-memory/blob/3d34260ae513af48030da9a56aa50b8e0162c6f8/extensions/Postgres/Postgres/Internals/PostgresDbClient.cs#L419

https://github.com/microsoft/kernel-memory/blob/3d34260ae513af48030da9a56aa50b8e0162c6f8/extensions/Postgres/Postgres/Internals/PostgresDbClient.cs#L432-L448

In particular, the `foreach` adds `sqlUserValues` (included `similarityPlaceholder= "@__min_similarity"`) to query parameters, but this parameter seems not to be used.

Instead, records are filtered by similarity after the query execution:

https://github.com/microsoft/kernel-memory/blob/3d34260ae513af48030da9a56aa50b8e0162c6f8/extensions/Postgres/Postgres/Internals/PostgresDbClient.cs#L458-L469

marcominerva:
> 
> But, wouldn't it be better & easier to read if the minSimilarity check would be part of the query? Then it could also be optimised by Postgres, instead of having an in-memory search / filtering afterwards. Would need to check `filterSql` a bit further, as in my test case it was empty / `TRUE`.

Yes, it is exactly what I mean when I said that `minSimilarity` parameter is not used in the query.

marcominerva:
What you have called `difference` is actually the **distance** (the closer it is to 0, the closer the vectors themselves are). I suggest you to rename the code in this way to better specify the intent.

Moreover, Kernel Memory works with the concept of **relevance** (the closer it is to 1, more relevant the vectors are). All memories return records with the most relevance first. See for example:

https://github.com/microsoft/kernel-memory/blob/3d34260ae513af48030da9a56aa50b8e0162c6f8/service/Core/Search/SearchClient.cs#L227-L238

Instead, now in you're code the [result list](https://github.com/microsoft/kernel-memory/blob/f791fb1b685d8d0773960e9db2ffc7bd26ad2214/extensions/Postgres/Postgres/Internals/PostgresDbClient.cs#L469) is ordered in an ascending way based on the **distance** (even if you calculate the similarity, the list is created reading the records that are returned by the query). So, I think you need to reverse the result list.

marcominerva:
> I have already reversed the ORDER BY from DESC (on similarity) to ASC (on distance), so I think that's correct. Upon further investigation it turns out the results were not different. Can you please do a review and see if it can be merged? @marcominerva

I too have made a test with the old and the new approach and I have verified that now results are returned in the correct order. I would suggest only a last minor fix: rename the `colDistance` variable using `__` as prefix, as in the original code:

`string colDistance = "__distance";`

Remember also to correct the comment at line 437 (you're still using `colDifference`).

After that, the code seems OK to me. Now me must wait for the approval by @dluc.

extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
```diff
@@ -447,24 +454,18 @@ OFFSET @offset
                     cmd.Parameters.AddWithValue(kv.Key, kv.Value);
                 }
 #pragma warning restore CA2100
-
+                this._log.LogTrace("SQL: {0}", cmd.CommandText);
```
dluc, line 457:
```suggestion
```

dluc:
could you grant maintainers edit on the PR?

extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
```diff
@@ -404,22 +404,24 @@ ON CONFLICT ({this._colId})
...
+        filterSql += $" AND {this._colEmbedding} <=> @embedding < {maxDistance}";
 
         if (sqlUserValues == null) { sqlUserValues = new(); }
 
-        sqlUserValues[similarityPlaceholder] = minSimilarity;
+        sqlUserValues[colMaxDistance] = minSimilarity;
 
         this._log.LogTrace("Searching by similarity. Table: {0}. Threshold: {1}. Limit: {2}. Offset: {3}. Using SQL filter: {4}",
-            tableName, minSimilarity, limit, offset, string.IsNullOrWhiteSpace(filterSql) ? "false" : "true");
+            tableName, minSimilarity, limit, offset, filterSql);
```
dluc, line 424: to revert: the filter cannot be logged because it could contain PII

extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
```diff
@@ -404,19 +404,21 @@ ON CONFLICT ({this._colId})
...
+        string colDistance = "__distance";
+        string colMaxDistance = "@__max_distance";
 
         // Filtering logic, including filter by similarity
         filterSql = filterSql?.Trim().Replace(PostgresSchema.PlaceholdersTags, this._colTags, StringComparison.Ordinal);
         if (string.IsNullOrWhiteSpace(filterSql))
         {
             filterSql = "TRUE";
         }
+        var maxDistance = 1 - minSimilarity;
```
dluc, line 416:
```suggestion

        var maxDistance = 1 - minSimilarity;
```

dluc: nit, codestyle: missing empty line

dluc:
> > could you grant maintainers edit on the PR?
> 
> Sorry @dluc , that option isn't there, I guess because it's not a user-owned fork, but an org owned? Unless I'm missing something, but I believe the option should appear here on the right at the bottom based on [Github Docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).

please upvote https://github.com/orgs/community/discussions/5634

dluc (approved):

PR #695, opened 2024-07-11
Add SslEnabled option for RabbitMQ adapter

koteus:
This option will add default SSL config in cases where RabbitMQ server uses SSL. Currently we can't connect to the most of remote RabbitMQ servers because of this, because secure connection is required.

Just one config is added `SslEnabled` which is boolean (false by default), when setting it to true, it will provide default SslOption to `RabbitMQ.Client`


tools/InteractiveSetup/Services/RabbitMQ.cs
```diff
@@ -33,6 +34,7 @@ public static void Setup(Context ctx, bool force = false)
             { "Username", SetupUI.AskOpenQuestion("RabbitMQ <username>", config["Username"].ToString()) },
             { "Password", SetupUI.AskPassword("RabbitMQ <password>", config["Password"].ToString()) },
             { "VirtualHost", SetupUI.AskOpenQuestion("RabbitMQ <virtualhost>", config["VirtualHost"].ToString()) },
+            { "SslEnabled", SetupUI.AskOpenQuestion("RabbitMQ <SSL enabled>", config["SslEnabled"].ToString()) },
```
dluc, line 37: should use `AskBoolean`, see for example how `UseHybridSearch` is set

dluc (changes requested):
LGTM, just one minor fix required before merging

koteus:
> LGTM, just one minor fix required before merging

✅ Done

dluc (approved):

PR #699, opened 2024-07-12
Add ChunkHeader option to request context

westdavidr:
I have a use case to prepend a header (in my case the original file path from sharepoint) in to each partition to be able to include that in the embedding and RAG. This is a minimal change with the addition of the RequestContext and leverages the existing `chunkHeader` parameter in the TextChunker.

Adds constant for `ChunkHeader` in the `Constants.CustomContext.Partitioning` class.

In `TextPartitioningHandler` we look for the existence of that context argument and default to `null` (existing behavior is null here)


westdavidr:

dluc (approved):

PR #717, opened 2024-07-25
Spike - add page number. Ask for OK to proceed.

alkampfergit:

I'd like to solve discussion #669 This is absolutely still not ready, but it is the starting point to close #669 

Introduced a new TextChunker that allows for specification of a series of line with a tag, then it will split based on token number, but instead of creating simple lines in output it will preserve the tag of the original line. I've chosen a tag object so we can add more info if needed, in this PR I've added the plain page number.

Needs to check if the overall implant of the  PR is ok, then I'll finish implementation and testing of TextChunk2.


service/Core/Handlers/TextPartitioningHandler.cs
dluc, line 1: Could you create a new handler, using the new chunker? I'd prefer keeping the two approaches separate, so we can move faster. The new handler can be registered like the others, or simply replace the old one tweaking the config file

alkampfergit: Fair enough, this will minimize the risk that the modification changes old behaviour.

alkampfergit: I've renamed the partitioning handler, I'm a little bit unsure of the name used, but this will minimize impact on existing code. 

I've a question, is there an example in the examples section that explain to the user how to change only the TextPartitioningHandler?

dluc:

PR #722, opened 2024-07-30
Add support for SQL Server Vector Search

marcominerva:
Azure SQL Database now provides Vector Support, currently in EAP (https://devblogs.microsoft.com/azure-sql/announcing-eap-native-vector-support-in-azure-sql-database). This PR extends SQL Server Memory DB with a flag that allows to use this new feature.

Just pass `true` to the new `useVectorSearch` argument:

```csharp
builder.Services.AddKernelMemory(options =>
{
    options
      //..
      .WithSqlServerMemoryDb("connecstion_string", useVectorSearch: true)
      //...
      ;
});
```

With this value, the Memory will use the new Vector Support. In particular, the table `KMEmbeddings_index` will no longer be necessary, because vectors are now stored in binary format and vector distance can be performed using directly the embedding column in the `KMMemories` table.

In other words, [SqlServerMemory](https://github.com/microsoft/kernel-memory/blob/main/extensions/SQLServer/SQLServer/SqlServerMemory.cs) will act differently based on the value of the `useVectorSearch` flag. I have chosen this approach, rather than creating a brand new **MemoryDb** implementation, because that are a lot of code in common. Of course, we can decide to split the implementation.

At this moment, vector dimension is limited to 1998, so we can use models like `text-embedding-ada-002` or `text-embedding-3-small`. In case of `text-embedding-3-large`, we need to specify the value of vector dimension when configuring Kernel Memory.

> [!IMPORTANT]  
> Remember that, at this time, Vector Support is available only on Azure SQL Database. On the other hand, the current SQL Server Memory DB requires a COLUMNSTORE INDEX that, on Azure, is available only on vCore databases and Standard databases in S3 and above pricing tiers (https://azure.microsoft.com/en-us/blog/columnstore-support-in-standard-tier-azure-sql-databases).


extensions/SQLServer/SQLServer/SqlServerConfig.cs
```diff
@@ -61,4 +61,9 @@ public class SqlServerConfig
     /// The SQL Server tags table name.
     /// </summary>
     public string TagsTableName { get; set; } = DefaultTagsTableName;
+
+    /// <summary>
+    /// It tells if we should use the vector search or not.
```
dluc, line 66: could you add a link to the relevant feature docs?

dluc:
I think the amount of SQL strings in the code is getting to a very risky state, hard to review for SQL injections (something we keep getting security warnings about). Would it be hard to refactor out all the SQL manipulation in a dependency, having one class for the old SQL and one for the new SQL with vector support?

extensions/SQLServer/SQLServer/SqlServerConfig.cs
marcominerva: The feature is currently in Early Adopter Preview, so the only useful link is this one: https://devblogs.microsoft.com/azure-sql/announcing-eap-native-vector-support-in-azure-sql-database. There isn't an official documentation page at the moment. Should we use this link, and update it with new resources when available?

marcominerva:
@dluc I have extracted all the SQL manipulations in external dependencies (see https://github.com/microsoft/kernel-memory/blob/0d81477a626e55c3f8dbc38461cfa884749f6a6e/extensions/SQLServer/SQLServer/QueryProviders/ISqlServerQueryProvider.cs).

Here it is how it is used:

https://github.com/microsoft/kernel-memory/blob/0d81477a626e55c3f8dbc38461cfa884749f6a6e/extensions/SQLServer/SQLServer/SqlServerMemory.cs#L71)

Currently, there are some code duplication in the two implementations, but for the moment I would like to know if it is the correct approach. Then, we could optimize the code.

extensions/SQLServer/SQLServer/SqlServerConfig.cs
dluc: yes the link would be useful :-)

dluc:
> @dluc I have extracted all the SQL manipulations in external dependencies (see https://github.com/microsoft/kernel-memory/blob/0d81477a626e55c3f8dbc38461cfa884749f6a6e/extensions/SQLServer/SQLServer/QueryProviders/ISqlServerQueryProvider.cs).
> 
> Currently, there are some code duplication in the two implementations, but for the moment I would like to know if it is the correct approach. Then, we could optimize the code.

yup, exactly what I was hoping for, thank you. Let me know when the PR is ready

marcominerva:
I have changed the `ISqlServerQueryProvider` interface to an [abstract base class](https://github.com/marcominerva/kernel-memory/blob/sqlserver-vectorsearch/extensions/SQLServer/SQLServer/QueryProviders/SqlServerQueryProvider.cs) so that it can contain the common logic. Now the PR is ready for the review ;)

KSemenenko:
What db do you use? For pgvector https://github.com/pgvector/pgvector-dotnet there is a client.

So it looks like it’s possible to use EntityFrameworkCore

extensions/SQLServer/SQLServer/DependencyInjection.cs
```diff
@@ -46,10 +47,10 @@ public static IKernelMemoryBuilder WithSqlServerMemoryDb(
 public static partial class DependencyInjection
 {
     /// <summary>
-    /// Inject SqlServer as the default implementation of IMemoryDb
+    /// Inject SQL Server as the default implementation of IMemoryDb
     /// </summary>
     /// <param name="services">Service collection</param>
-    /// <param name="config">Postgres configuration</param>
+    /// <param name="config">SQL Service configuration</param>
```
dluc, line 53:
```suggestion
    /// <param name="config">SQL Server configuration</param>
```

dluc (commented):
review in progress...

marcominerva:
Hi @dluc! The native Vector type for SQL Database has just been announced: https://devblogs.microsoft.com/azure-sql/eap-for-vector-support-refresh-introducing-vector-type/. As soon as I get acces to it, I'll update this PR with the new support.

dluc (approved):

Issue #679, opened 2024-06-26
[Bug] Connection pool not released after Postgres query


xuzeyu91:
### Context / Scenario

```
await memoryDb.GetListAsync(KmsConstantcs.KmsIndex, filters: new List<MemoryFilter>() { new MemoryFilter().ByDocument(fileId) }, limit: int.MaxValue).ToListAsync().ConfigureAwait(false);
```
I often encounter anomalies such as sorry, too many clients are already ready when querying document slices. I have observed that each query adds a new link

```
SELECT count(*) FROM pg_stat_activity;
```
I have found that every query adds new links and they are not dispose. Frequent queries will cause my Postgres connection pool to fill up

### What happened?

too many clients are already ready

### Importance

I cannot use Kernel Memory

### Platform, Language, Versions

I use kernel-memory to connect to Postgres

### Relevant log output

```shell
too many clients are already ready
```

xuzeyu91:
https://github.com/microsoft/kernel-memory/blob/3d34260ae513af48030da9a56aa50b8e0162c6f8/extensions/Postgres/Postgres/Internals/PostgresDbClient.cs#L529

I'm not sure if there's a problem here

xuzeyu91:
```
 _memoryDbs = memory.Orchestrator.GetMemoryDbs();
 foreach (var memoryDb in _memoryDbs)
 {
     var items = await memoryDb.GetListAsync(KmsConstantcs.KmsIndex, filters: new List<MemoryFilter>() { new MemoryFilter().ByDocument(fileId) }, limit: int.MaxValue).ToListAsync().ConfigureAwait(false);
 }
```
This is my code, I need to retrieve the slice data from the document, which will cause the number of connections to increase continuously

marcominerva:
@dluc, could it be related to the fact that `PostgresMemory` is registered as singleton? `PostgresMemory` in turns creates a `PostgresDbClient`. So, even if both `PostgresMemory` and `PostgresDbClient` implements `IDisposable`, the `Dispose` methods are never called.

In particular, I see that `PostgresDbClient` contains an `NpgsqlDataSource` object that is created in the constructor but never disposed:
https://github.com/microsoft/kernel-memory/blob/3d34260ae513af48030da9a56aa50b8e0162c6f8/extensions/Postgres/Postgres/Internals/PostgresDbClient.cs#L31

dluc:
Each connection is opened and released using this pattern:

```csharp
[...]

PR #771, opened 2024-09-05
Add disposal of NpgsqlDataSource after opening connection

marcominerva:
See https://github.com/microsoft/kernel-memory/issues/679.

If I call `ImportDocumentAsync` and then `AskAsync`, after the execution I see that two sessions are still in the list:

image

This PR adds the disposal the **NpgsqlDataSource** after each connection. In this way, no pending sessions remain when the execution completes.

~~NOTE: for the moment, I have left the `IDisposable` interface on **PostgresDbClient.cs** (with an empty implementation for reference), but actually it is no longer necessary, as explained in https://github.com/microsoft/kernel-memory/issues/679#issuecomment-2202303865. So, I think it could be safely removed.~~

### DIFF
--- a/extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
+++ b/extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
@@ -19,7 +19,7 @@ namespace Microsoft.KernelMemory.Postgres;
 /// <summary>
 /// An implementation of a client for Postgres. This class is used to managing postgres database operations.
 /// </summary>
-internal sealed class PostgresDbClient : IDisposable
+internal sealed class PostgresDbClient
 {
     // See: https://www.postgresql.org/docs/current/errcodes-appendix.html
     private const string PgErrUndefinedTable = "42P01"; // undefined_table
@@ -229,7 +229,7 @@ internal sealed class PostgresDbClient : IDisposable
             {
                 // Check if the custom SQL contains the lock placeholder (assuming it's not commented out)
                 bool missingLockStatement = !string.IsNullOrEmpty(this._createTableSql)
-                                             && !this._createTableSql.Contains(PostgresConfig.SqlPlaceholdersLockId, StringComparison.Ordinal);
+                    && !this._createTableSql.Contains(PostgresConfig.SqlPlaceholdersLockId, StringComparison.Ordinal);
 
                 if (missingLockStatement)
                 {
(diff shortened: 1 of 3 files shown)


extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
```diff
@@ -685,7 +684,8 @@ private async Task<NpgsqlConnection> ConnectAsync(CancellationToken cancellation
     {
         try
         {
-            return await this._dataSource.OpenConnectionAsync(cancellationToken).ConfigureAwait(false);
+            using var dataSource = this._dataSourceBuilder.Build();
```
dluc, line 687: should use the async dispose right? ie `await using`

extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
```diff
@@ -28,7 +28,7 @@ internal sealed class PostgresDbClient : IDisposable
     private const string PgErrDatabaseDoesNotExist = "3D000"; // invalid_catalog_name
 
     private readonly ILogger _log;
-    private readonly NpgsqlDataSource _dataSource;
+    private readonly NpgsqlDataSourceBuilder _dataSourceBuilder;
```
dluc, line 31: by removing `NpgsqlDataSource` we can also remove `IDisposable` (line 22) and the `Dispose` methods

extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
```diff
@@ -228,8 +228,8 @@ public async Task CreateTableAsync(
             if (await this.DoesTableExistAsync(origInputTableName, cancellationToken).ConfigureAwait(false))
             {
                 // Check if the custom SQL contains the lock placeholder (assuming it's not commented out)
-                bool missingLockStatement = (!string.IsNullOrEmpty(this._createTableSql)
-                                             && !this._createTableSql.Contains(PostgresConfig.SqlPlaceholdersLockId, StringComparison.Ordinal));
+                bool missingLockStatement = !string.IsNullOrEmpty(this._createTableSql)
+                                             && !this._createTableSql.Contains(PostgresConfig.SqlPlaceholdersLockId, StringComparison.Ordinal);
```
dluc, line 232: nit: indentation

dluc (changes requested):

extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
marcominerva: I have removed the `IDsposable` interface from `NpgsqlDataSource `, that in turn makes the `IDisposable` interface on `PostgresMemory` on longer necessary.

dluc (approved):

PR #788, opened 2024-09-17
Added initial implementation of Microsoft.KernelMemory.AI.ONNX package

JustinRidings:
This change includes an entirely new extension that allows TextGeneration with ONNX Models. This scenario allows users to interact with local cloned ONNX Models

This is a scenario that abstracts the general implementation of the OnnxRuntimeGenAI library to work with text generation. More examples on using OnnxGenAI Libraries, please see https://github.com/microsoft/onnxruntime-genai


extensions/ONNX/Onnx/OnnxConfig.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// </summary>
+    public uint MaxLength { get; set; } = 2048;
+
+    /// <summary>
+    /// The minimum length of the response that the model will generate. See https://onnxruntime.ai/docs/genai/reference/config.html
+    /// </summary>
+    public uint MinLength { get; set; } = 0;
+
+    /// <summary>
+    /// The option to use a share one buffer for past and present for efficiency. See https://onnxruntime.ai/docs/genai/reference/config.html
```
dluc, line 24: I checked the docs but I could not understand what this means. If the buffer is shared across two requests, is there a risk that two different users will interfere with each other, risking also to leak personal data?

extensions/ONNX/Onnx/OnnxConfig.cs
```diff
@@ -0,0 +1,43 @@
...
+    public void Validate(bool allowIO = true)
+    {
+        if (string.IsNullOrWhiteSpace(this.ModelPath))
+        {
+            throw new ConfigurationException($"Onnx: {nameof(this.ModelPath)} is empty");
+        }
+
+        if (allowIO && !File.Exists(this.ModelPath))
+        {
+            throw new ConfigurationException($"Onnx: {nameof(this.ModelPath)} file not found");
```
dluc, line 40: I would include the path rather than "ModelPath" in this err msg - wdyt?

extensions/ONNX/Onnx/OnnxConfig.cs
```diff
@@ -0,0 +1,43 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+#pragma warning disable IDE0130 // reduce number of "using" statements
+// ReSharper disable once CheckNamespace - reduce number of "using" statements
+namespace Microsoft.KernelMemory;
```
dluc, line 5:
```suggestion
namespace Microsoft.KernelMemory;

```

dluc: (missing empty line)

extensions/ONNX/Onnx/OnnxTextGenerator.cs
```diff
@@ -0,0 +1,138 @@
...
+        string prompt,
+        TextGenerationOptions? options = null,
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    {
+        var tokens = this._tokenizer?.Encode(prompt);
+        using var generatorParams = new GeneratorParams(this._model);
+
+        generatorParams.SetSearchOption("max_length", (double)this.MaxTokenTotal);
+        generatorParams.SetSearchOption("min_length", this.MinLength);
+        generatorParams.SetSearchOption("past_present_share_buffer", this.PastPresentShareBuffer);
```
dluc, line 79: I think the code is missing settings for `top_p` (nucleus sampling), `temperature`, `repetition_penalty`, maybe others?

extensions/ONNX/Onnx/OnnxTextGenerator.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        if (options != null)
+        {
+            generatorParams.SetSearchOption("temperature", options.Temperature);
+        }
+
+        await Task.Run(() => generatorParams.SetInputSequences(tokens), cancellationToken).ConfigureAwait(true);
```
dluc, line 86: `generatorParams` could be disposed outside

why the need to run a task this way?

extensions/ONNX/Onnx/OnnxTextGenerator.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        using (var generator = new Generator(this._model, generatorParams))
+        {
+            List<int> outputTokens = new();
```
dluc, line 90: maybe I'm missing something, is this list updated somewhere?

extensions/ONNX/Onnx/OnnxTextGenerator.cs
```diff
@@ -0,0 +1,138 @@
...
+    /// <inheritdoc/>
+    public IReadOnlyList<string> GetTokens(string text)
+    {
+        return this._textTokenizer.GetTokens(text);
+    }
+
+    /// <inheritdoc/>
+    public void Dispose()
+    {
+        this.Dispose();
```
dluc, line 121: the method is calling itself

extensions/ONNX/Onnx/OnnxTextGenerator.cs
```diff
@@ -0,0 +1,138 @@
...
+        GC.SuppressFinalize(this);
+    }
+
+    /// <inheritdoc/>
```
dluc, line 125:
```suggestion
```

extensions/ONNX/README.md
```diff
@@ -0,0 +1,25 @@
...
+This project contains the
+[ONNX](https://onnxruntime.ai/docs/genai/)
+LLM connector to access to LLM models via Onnx service to generate text.
+
+Sample code:
+
+```csharp
+var config = new OnnxConfig
+{
+    ModelPath = "C:\\..\\Phi-3-mini-128k-instruct-onnx\\..\\cpu-int4-rtn-block-32"
```
dluc, line 15: nit: replace `..` with `....`, or it looks like "parent folder" :-)

extensions/ONNX/Onnx/Onnx.csproj
dluc, line 1: for consistency, if we're using `Onnx` I would rename the parent `ONNX` folder to `Onnx` too

extensions/ONNX/Onnx/OnnxConfig.cs
JustinRidings: Will reach out to the OnnxRuntimeGenAi team about this one, but as mitigation to the security concern you have, I'll be removing this modeled concept until I get more clarity.

extensions/ONNX/Onnx/OnnxTextGenerator.cs
JustinRidings: Using a Task.Run to remove the compiler warning about the method not being async without an await. 

Removing async from the method entirely throw [Error CS8403](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs8403?f1url=%3FappId%3Droslyn%26k%3Dk(CS8403))

extensions/ONNX/Onnx/OnnxTextGenerator.cs
JustinRidings: Good catch 🥇

extensions/ONNX/Onnx/OnnxConfig.cs
```diff
@@ -0,0 +1,143 @@
...
+#pragma warning disable IDE0130 // reduce number of "using" statements
+
+namespace Microsoft.KernelMemory;
+
+public class OnnxConfig
+{
+    /// <summary>
+    /// Path to the directory containing the .ONNX file.
+    /// </summary>
+    public string ModelPath { get; set; } = string.Empty;
```
dluc, line 12: Do you think we should allow using different models for text generation and embeddings? 

For instance, I would like to use E5 for embeddings and Phi3.5 for text. 

Thoughts?

dluc: see for example: https://github.com/microsoft/kernel-memory/blob/main/extensions/Ollama/Ollama/OllamaConfig.cs

extensions/ONNX/Onnx/OnnxTextGenerator.cs
```diff
@@ -0,0 +1,178 @@
...
+/// Text generator based on Onnx models, via OnnxRuntimeGenAi
+/// See https://github.com/microsoft/onnxruntime-genai
+/// </summary>
+[Experimental("KMEXP01")]
+public sealed class OnnxTextGenerator : ITextGenerator, IDisposable
+{
+    private readonly Model? _model = default;
+    private readonly Tokenizer? _tokenizer = default;
+    private readonly ILogger<OnnxTextGenerator> _log;
+    private readonly ITextTokenizer _textTokenizer;
```
dluc, line 23: could you add some docs on _tokenizer and _textTokenizer? even after reading the code, it's not clear why there are two

extensions/ONNX/Onnx/OnnxTextGenerator.cs
```diff
@@ -0,0 +1,178 @@
...
+        this._model = new Model(config.ModelPath);
+        this._tokenizer = new Tokenizer(this._model);
+        this._log.LogDebug("Onnx model loaded");
+    }
+
+    internal OnnxConfig _config { get; } = new OnnxConfig();
+
+    public int MaxTokenTotal { get; internal set; }
+
+    public double Temperature { get; internal set; }
```
dluc, line 64: nit: move these lines before the ctor

extensions/ONNX/Onnx/OnnxTextGenerator.cs
```diff
@@ -0,0 +1,178 @@
...
+        using (var generator = new Generator(this._model, generatorParams))
+        {
+            List<int> outputTokens = new();
+
+            while (!generator.IsDone() && cancellationToken.IsCancellationRequested == false)
+            {
+                generator.ComputeLogits();
+                generator.GenerateNextToken();
+
+                await Task.Run(() => outputTokens.AddRange(generator.GetSequence(0)), cancellationToken).ConfigureAwait(true);
```
dluc, line 135: I would suggest using a `await Task.CompletedTask.ConfigureAwait(false);` at the end of the method instead, to avoid using multiple tasks here.

extensions/ONNX/Onnx/OnnxTextGenerator.cs
```diff
@@ -0,0 +1,178 @@
...
+        this._textTokenizer = textTokenizer;
+
+        var modelFilename = config.ModelPath.TrimEnd('/').Split('/').Last();
+        this._log.LogDebug("Loading Onnx model: {0}", modelFilename);
+        this._model = new Model(config.ModelPath);
+        this._tokenizer = new Tokenizer(this._model);
+        this._log.LogDebug("Onnx model loaded");
+    }
+
+    internal OnnxConfig _config { get; } = new OnnxConfig();
```
dluc, line 60: this can be a `private` field

extensions/ONNX/Onnx/OnnxConfig.cs
```diff
@@ -0,0 +1,143 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+#pragma warning disable IDE0130 // reduce number of "using" statements
+
+namespace Microsoft.KernelMemory;
```
dluc, line 5:
```suggestion
#pragma warning disable IDE0130 // reduce number of "using" statements
// ReSharper disable once CheckNamespace - reduce number of "using" statements
namespace Microsoft.KernelMemory;
```

to fix warnings

extensions/ONNX/Onnx.FunctionalTests/OnnxTextGeneratorTest.cs
```diff
@@ -0,0 +1,65 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Diagnostics;
+using System.Text;
+using Microsoft.KernelMemory.AI;
+using Microsoft.KernelMemory.AI.Onnx;
+using Microsoft.KernelMemory.AI.OpenAI;
```
dluc, line 7:
```suggestion
using Microsoft.KernelMemory.AI.OpenAI;
```

not used

dluc (changes requested):

extensions/ONNX/Onnx/OnnxTextGenerator.cs
JustinRidings: Thats a clever way to implement an async :)

dluc (approved):

PR #808, opened 2024-09-27
Make SK text and embedding generator wrappers public

dluc:
Currently it's possible to inject Semantic Kernel classes into KM only using 

- IKernelMemoryBuilder.WithSemanticKernelTextGenerationService
- IKernelMemoryBuilder.WithSemanticKernelTextEmbeddingGenerationService
- IServiceCollection.AddSingleton<ITextGenerator>(instance)
- IServiceCollection.AddSingleton<ITextEmbeddingGenerator>(instance)

However, it's not possible to create an instance of Semantic Kernel classes directly without writing a custom wrapper class first. Since the project already includes such wrapper, as internal classes, this PR changes them to public classes, to reduce the amount of code to write.

Classes:
- SemanticKernelTextGenerator
- SemanticKernelTextEmbeddingGenerator


dluc:
See also https://github.com/microsoft/kernel-memory/pull/801

PR #860, opened 2024-10-23
Add ability to control MaxMatchesCount for Ask via IContext

westdavidr:
Need the ability to control the MaxMatchesCount per Ask request instead of always defaulting to the configuration value.

Add  MaxMatchesCount constant and extensions.


westdavidr:
@dluc Let me know what you think of this. I have a need to control the max matches per request (generally per index, but that's controlled from the client logic) when using the /Ask endpoint.

dluc (approved):
LGTM 👍

PR #867, opened 2024-10-30
Fix embeddings not being returned in Azure AI Search memory

alexmg:

The `withEmbeddings` parameter for the `GetSimilarListAsync` and `GetListAsync` methods on the `AzureAISearchMemory` class are not being applied because the `vector` field in the search index is not retrievable.

Ensure that the index is created with the vector field as retrievable so the option can be applied.

- Add a `VectorSearchField` to the schema with the `IsHidden` and `IsStored` property values set
- Create a private method that sets the `Select` fields on the `SearchOptions` to ensure only needed fields are returned


dluc:
Comparing old and new schema.

Before:

image

After:

image

Diff shows only 1 change 👍 

image

dluc (approved):

PR #870, opened 2024-10-30
Refactor DataPipelineStatus and DataPipeline classes

marcominerva:
Removed the `Failed` property from `DataPipelineStatus` and reordered the `Empty` property with `JsonPropertyOrder` of 1. Updated `DataPipeline` class: refactored `Log` method to use null-coalescing assignment and adjusted `ToDataPipelineStatus` method to reflect the removal of the `Failed` property.

See https://github.com/microsoft/kernel-memory/discussions/709

### DIFF
--- a/service/Abstractions/Models/DataPipelineStatus.cs
+++ b/service/Abstractions/Models/DataPipelineStatus.cs
@@ -15,7 +15,7 @@ public class DataPipelineStatus
     /// <summary>
     /// Currently used to track deleted documents.
     /// </summary>
-    [JsonPropertyOrder(1)]
+    [JsonPropertyOrder(2)]
     [JsonPropertyName("empty")]
     public bool Empty { get; set; } = false;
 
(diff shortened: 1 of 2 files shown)


service/Abstractions/Models/DataPipelineStatus.cs
```diff
@@ -12,15 +12,10 @@ public class DataPipelineStatus
...
-    [JsonPropertyOrder(1)]
-    [JsonPropertyName("failed")]
-    public bool Failed { get; set; } = false;
-
     /// <summary>
     /// Currently used to track deleted documents.
-    /// TODO: replace with "isDeleting" and "Deleted"
     /// </summary>
-    [JsonPropertyOrder(2)]
+    [JsonPropertyOrder(1)]
```
dluc, line 18:
```suggestion
    [JsonPropertyOrder(2)]
```

we can leave this unchanged, no impact

PR #872, opened 2024-10-30
Add token usage tracking

marcominerva:
This PR adds a new `TokenUsage` property to `MemoryAnswer` to hold information about token usage.

Token usage is calculated in the `SearchClient.AskAsync` method using the configured tokenizer.

### DIFF
--- a/service/Abstractions/Models/MemoryAnswer.cs
+++ b/service/Abstractions/Models/MemoryAnswer.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using System.Text;
 using System.Text.Json;
 using System.Text.Json.Serialization;
+using Microsoft.KernelMemory.Models;
 
 namespace Microsoft.KernelMemory;
 
@@ -40,6 +41,14 @@ public class MemoryAnswer
     [JsonPropertyName("text")]
     [JsonPropertyOrder(10)]
     public string Result { get; set; } = string.Empty;
+
+    /// <summary>
+    /// The tokens used by the model to generate the answer.
+    /// </summary>
+    /// <remarks>Not all the models and text generators return token usage information.</remarks>
+    [JsonPropertyName("tokenUsage")]
+    [JsonPropertyOrder(11)]
+    public TokenUsage TokenUsage { get; set; } = new();
 
     /// <summary>
     /// List of the relevant sources used to produce the answer.
--- a/service/Abstractions/Models/TokeUsage.cs
+++ b/service/Abstractions/Models/TokeUsage.cs
@@ -1,0 +1,29 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Text.Json.Serialization;
+
+namespace Microsoft.KernelMemory.Models;
+
+/// <summary>
+/// Represents the usage of tokens in a request and response cycle.
+/// </summary>
+public class TokenUsage
+{
+    /// <summary>
+    /// The number of tokens in the request message input, spanning all message content items.
+    /// </summary>
+    [JsonPropertyOrder(0)]
+    public int InputTokenCount { get; set; }
+
+    /// <summary>
+    /// The combined number of output tokens in the generated completion, as consumed by the model.
+    /// </summary>
+    [JsonPropertyOrder(1)]
+    public int OutputTokenCount { get; set; }
+
+    /// <summary>
+    /// The total number of combined input (prompt) and output (completion) tokens used.
+    /// </summary>
+    [JsonPropertyOrder(2)]
+    public int TotalTokenCount => this.InputTokenCount + this.OutputTokenCount;
+}
--- a/service/Core/Search/SearchClient.cs
+++ b/service/Core/Search/SearchClient.cs
@@ -336,8 +336,8 @@ public sealed class SearchClient : ISearchClient
             return noAnswerFound;
         }
 
-        var (prompt, tokenCount) = this.CreatePrompt(question, facts.ToString(), context);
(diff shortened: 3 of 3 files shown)


dluc:
I think we should use actual usage reports from the AI services when available, to avoid the risk of returning incorrect information, particularly when token count is important for billing. 

I would also design the report to support a list, because multiple AI calls are involved, and potentially to different models and different services.

For models where the report data is unavailable, we can fallback to tokenizers, using different keys like `estimated_input_tokens` for example.

Something like:

```json
{
  "requests": [
    {
      "id": "A0E4C1D0-0D1A-4D3B-8D3D-3D0D1A0E4C1D",
      "date": "2024-10-30T12:00:00Z",
      "service": "Azure OpenAI",
      "model": "GPT-4o-mini",
      "usage": {
        "input_tokens": 123,
        "output_tokens": 456,
        "output_reasoning_tokens": 50
      }
    },
    {
      "id": "C1D0B1E4-4D3B-0D1A-8D3D-E4C1D3D0D1A0",
      "date": "2024-10-30T12:00:00Z",
      "service": "Azure OpenAI",
      "model": "text-embedding-ada-002",
      "usage": {
        "prompt_tokens": 123,
        "total_tokens": 123
      }
    },
    {
      "date": "2024-10-30T12:00:00Z",
      "service": "LlamaSharp",
      "model": "llama2",
      "usage": {
        "estimated_input_tokens": 123,
        "estimated_output_tokens": 456
      }
    }
  ]
}
```

marcominerva:
> I think we should use actual usage reports from the AI services when available
> 

In fact, I initially started with that kind of implementation in my branch https://github.com/marcominerva/kernel-memory/tree/token_usage (see in particular https://github.com/marcominerva/kernel-memory/blob/5a3a77f62a2a22d88fe85ad5efb8426731e9d4a5/extensions/AzureOpenAI/AzureOpenAITextGenerator.cs#L148-L150), but then I found some blocking issues like https://github.com/microsoft/semantic-kernel/issues/9420.

So, I thought that we can start with the "manual" approach, and then progressively update it with the actual usage reports from the different services.

Token usage is one of the most request feature by my customers, and having a value such the one proposed in this PR, even if not 100% accurate, is much better than not having a value at all 😄 .  Also because, at the moment, they use this exact approach when they need to get an idea of token usage, with the problem that the actual prompt is "hidden", and the only point in which its size is shown is in the logger:

https://github.com/microsoft/kernel-memory/blob/67472d523c7c02e41938d8dd974bbe1620254456/service/Core/Search/SearchClient.cs#L409-L416

dluc:
> > I think we should use actual usage reports from the AI services when available
> 
> In fact, I initially started with that kind of implementation in my branch https://github.com/marcominerva/kernel-memory/tree/token_usage (see in particular https://github.com/marcominerva/kernel-memory/blob/5a3a77f62a2a22d88fe85ad5efb8426731e9d4a5/extensions/AzureOpenAI/AzureOpenAITextGenerator.cs#L148-L150), but then I found some blocking issues like [microsoft/semantic-kernel#9420](https://github.com/microsoft/semantic-kernel/issues/9420).
> 
> So, I thought that we can start with the "manual" approach, and then progressively update it with the actual usage reports from the different services.
> 
> Token usage is one of the most request feature by my customers, and having a value such the one proposed in this PR, even if not 100% accurate, is much better than not having a value at all 😄 . Also because, at the moment, they use this exact approach when they need to get an idea of token usage, with the problem that the actual prompt is "hidden", and the only point in which its size is shown is in the logger:
> 
> https://github.com/microsoft/kernel-memory/blob/67472d523c7c02e41938d8dd974bbe1620254456/service/Core/Search/SearchClient.cs#L409-L416

I would use the service data when available and include the tokenizer optionally. Using the tokenizer is a performance concern, so it would be nice if we could turn it off.

Something like this:

### PR 1

* applies only to OpenAI generators, text and embeddings 
* define a common class for metrics, avoiding dictionaries
* add token metrics to Search and Ask responses, organized as a list of calls
* each call includes
    * timestamp
    * name of the service e.g. "Azure OpenAI", "Ollama", etc.
    * model type e.g. "TextEmbedding", "Text", etc.
    * name of the model/deployment used
    * "tokens in metric" reported by the service - NULL if not available (key: "service_tokens_in")
    * "tokens out metric" reported by the service - NULL if not available (key: "service_tokens_out")
    * other token metrics reported by the service if avaialvble, e.g. reasoning tokens (key: "service_reasoning_tokens")
    * "token in" measured by the tokenizer (key: "tokenizer_tokens_in")
    * "token out" measured by the tokenizer (key: "tokenizer_tokens_out")

### PR 2  

* apply to all AI generators: Azure OpenAI, Ollama, LLamaSharp, Anthropic
* add global option "enable token metrics"
* add global option "include token metrics measured by tokenizer" - enabled by default

### PR 3 

* add report also to the pipeline object, to cost ingestion. Support updates, don't lose previous metrics.

nurkmez2:
Any updates? 👍

marcominerva:
I plan to continue working on this feature next week.

nurkmez2:
Hi @marcominerva 
Any updates? 
Thanks

marcominerva:
Sorry, unfortunately in these weeks I had a lot of unexpected works, so I had to delay this activity. I hope to resume the activity soon.

marcominerva:
@nurkmez2 I have made a new PR with the approach suggested by @dluc: https://github.com/microsoft/kernel-memory/pull/947. I think that this one can be safely closed.

PR #877, opened 2024-11-04
Add setting to enable SessionId on SearchOptions in AzureAISearchMemory

alexmg:

The `UseSessionId` property on the `SearchOptions` can help improve the relevance of search results when multiple replicas are configured. This should be the case in any production deployment as two replicas are required for high availability of read-only workloads (queries), and three or more for high availability of read-write workloads (queries and indexing).

> A value to be used to create a sticky session, which can help getting more consistent results. As long as the same sessionId is used, a best-effort attempt will be made to target the same replica set. Be wary that reusing the same sessionID values repeatedly can interfere with the load balancing of the requests across replicas and adversely affect the performance of the search service. The value used as sessionId cannot start with a '_' character.

https://learn.microsoft.com/en-us/rest/api/searchservice/documents/search-post?view=rest-searchservice-2024-07-01&tabs=HTTP#request-body

Using this setting should also provide more consistent results from the `AzureAISearchMemory.GetListAsync` method. Without an `orderby` parameter results are sorted by score, and when `*` is used as the search query, all results have the same score of `1`. This increases the likelihood of getting inconsistent results when the search query hits different replicas.

Extend the `AzureAISearchConfig` class with a new `UseSessionId` property:

- Default the setting to `false` to avoid any change in behaviour for existing users and to make its use an explicit choice like `UseHybridSearch`
- Rename the `DefineFieldsToSelect` method in `AzureAISearchMemory` to `ApplyCommonSearchOptions` as it will be used to also apply the `UseSessionId` setting
- When the setting is `true` set the `SessionId` property on the `SearchOptions` to `Guid.NewGuid().ToString("N")` ensuring the same session ID is used while paging results but is not reused for subsequent method calls
- Take the opportunity to remove duplicate code present for setting the `SearchOptions.Filter` property by passing the `ICollection<MemoryFilter>?` to the `ApplyCommonSearchOptions` method
- Update the interactive setup to include a question to set the `UseSessionId` property
- Include the new setting with comment in the various `appsettings.json` files in the solution


dluc (approved):
Nice add!

dluc (approved):

PR #883, opened 2024-11-08
Bugfix SQLServerMemory semaphore not released properly

snakex64:

A simple bugfix for SQLServerMemory. The semaphore wasn't released when "is ready" changes from "false" to "true" while the code was waiting for the semaphore. Meaning if specifically 3 or more tasks run into that code at the same time, the third task as well as all others are gonna be stuck in a deadlock waiting for the semaphore. This only happens during the first initialization, since subsequent calls will leave before waiting for the semaphore.

### DIFF
--- a/extensions/SQLServer/SQLServer/SqlServerMemory.cs
+++ b/extensions/SQLServer/SQLServer/SqlServerMemory.cs
@@ -397,10 +397,11 @@ public sealed class SqlServerMemory : IMemoryDb, IMemoryDbUpsertBatch, IDisposab
         if (this._isReady) { return; }
 
         await this._initSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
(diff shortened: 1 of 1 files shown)


dluc (approved):

marcominerva:
@snakex64 @dluc Surely I'm missing something, but I'm wondering how this change (an `if` instrctuion moved inside a `try... catch` block):

image

Will modify the behavior?

dluc:
> @snakex64 @dluc Surely I'm missing something, but I'm wondering how this change (an `if` instrctuion moved inside a `try... catch` block):
> 
> image
> 
> Will modify the behavior?

it's about the code executed in `finally`

marcominerva:
Oh... I see, in this way the `return` causes the semaphore to be released. Thank you for the clarification!

PR #892, opened 2024-11-15
Update README.md

imranshams:
Fixed incorrect link for swagger

The swagger link is incorrect and I updated

Just the link correction


dluc (approved):
thanks!

PR #893, opened 2024-11-15
Fix Max Upload Size handling

marcominerva:
See https://github.com/microsoft/kernel-memory/pull/863#issuecomment-2443342730 and the comments below

### DIFF
--- a/service/Core/Configuration/ServiceConfig.cs
+++ b/service/Core/Configuration/ServiceConfig.cs
@@ -30,18 +30,8 @@ public class ServiceConfig
     public Dictionary<string, HandlerConfig> Handlers { get; set; } = new();
 
     /// <summary>
-    /// The maximum allowed size in megabytes for a request body posted to the upload endpoint.
+    /// The maximum allowed size in bytes for a request body posted to the upload endpoint.
     /// If not set the solution defaults to 30,000,000 bytes (~28.6 MB) (ASP.NET default).
     /// </summary>
-    public long? MaxUploadSizeMb { get; set; } = null;
-
-    public long? GetMaxUploadSizeInBytes()
-    {
-        if (this.MaxUploadSizeMb.HasValue)
-        {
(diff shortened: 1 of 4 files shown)


service/Core/Configuration/ServiceConfig.cs
```diff
@@ -30,18 +29,8 @@ public class ServiceConfig
     public Dictionary<string, HandlerConfig> Handlers { get; set; } = new();
 
     /// <summary>
-    /// The maximum allowed size in megabytes for a request body posted to the upload endpoint.
+    /// The maximum allowed size in bytes for a request body posted to the upload endpoint.
```
dluc, line 32: as per conversation, I'd prefer sticking to "megabytes"

service/Core/Configuration/ServiceConfig.cs
```diff
@@ -30,18 +29,8 @@ public class ServiceConfig
...
     /// If not set the solution defaults to 30,000,000 bytes (~28.6 MB) (ASP.NET default).
     /// </summary>
-    public long? MaxUploadSizeMb { get; set; } = null;
-
-    public long? GetMaxUploadSizeInBytes()
-    {
-        if (this.MaxUploadSizeMb.HasValue)
-        {
-            return Math.Min(10, this.MaxUploadSizeMb.Value) * 1024 * 1024;
```
dluc, line 42: I think this is the only line that needs fixing, using `Math.Max`. I'd keep also the min of 10Mb in place.

dluc (changes requested):
see comments

service/Core/Configuration/ServiceConfig.cs
marcominerva: So don't you want to handle scenarios where, for example, I need to limit the upload to 1 MB?

marcominerva:
I think that this PR has been superseded by https://github.com/microsoft/kernel-memory/pull/896, so it can be closed.

PR #912, opened 2024-11-27
Allow to select tokenizers via configuration

dluc:

Allow to configure tokenizers like any other dependency, removing several instances of warnings appearing in logs and using most common defaults.
When using OpenAI try to autodetect the correct tokenizer by model name, reducing the needed config.

- allow to configure tokenizer dependencies
- auto detect tokenizer when using OpenAI
- fix a few DI missing params


marcominerva:
Hi @dluc!

I have a question about this PR. In **OpenAIConfig.cs** file there is this new property:

https://github.com/microsoft/kernel-memory/blob/53db61a57e8aeafe6a5b8e6aec9cde11c401313a/extensions/OpenAI/OpenAI/OpenAIConfig.cs#L55-L59

The corresponding property in **AzureOpenAIConfig.cs** is the following:

https://github.com/microsoft/kernel-memory/blob/53db61a57e8aeafe6a5b8e6aec9cde11c401313a/extensions/AzureOpenAI/AzureOpenAI/AzureOpenAIConfig.cs#L66-L70

So, in the first case the comment says _Leave empty for autodetect_, in fact the default value of the property is `string.Empty`. On the other hand, in **AzureOpenAIConfig.cs** it is said _Leave it empty if unsure_, but the default value is `cl100k`. Is it intended or is it a typo?

dluc:
> Hi @dluc!
> 
> I have a question about this PR. In **OpenAIConfig.cs** file there is this new property:
> 
> https://github.com/microsoft/kernel-memory/blob/53db61a57e8aeafe6a5b8e6aec9cde11c401313a/extensions/OpenAI/OpenAI/OpenAIConfig.cs#L55-L59
> 
> The corresponding property in **AzureOpenAIConfig.cs** is the following:
> 
> https://github.com/microsoft/kernel-memory/blob/53db61a57e8aeafe6a5b8e6aec9cde11c401313a/extensions/AzureOpenAI/AzureOpenAI/AzureOpenAIConfig.cs#L66-L70
> 
> So, in the first case the comment says _Leave empty for autodetect_, in fact the default value of the property is `string.Empty`. On the other hand, in **AzureOpenAIConfig.cs** it is said _Leave it empty if unsure_, but the default value is `cl100k`. Is it intended or is it a typo?

hi @marcominerva, the comments are intentional - no typos.

When working with OpenAI, we can determine the appropriate tokenizer automatically by checking the model name. The only exception that I'm aware of, is when dealing with fine-tuned models, where we can rely on the developer to configure accordingly (maybe we should mention this in the comment).

With Azure, however, we only have a "deployment name" to work with, so the configuration defaults are designed to minimize errors and avoid warnings in the logs. Since all OpenAI embedding models use cl100k, I considered it a reasonable default. For gpt-4o text models, o200k is technically used, but the differences are minimal. While not 100% precise, this approach should avoid significant issues.

If this default causes problems, I’m open to suggestions. I did consider autodetecting the tokenizer by calling the API, but that felt overly complex and somehow expensive. Perhaps a separate tool for this task could be explored. That said, the lack of precision affects other models as well, like Llama and Anthropic, where we’re also using OpenAI tokenizers. Given this broader context, I think this implementation is reasonable, but I’m happy to hear other perspectives.

marcominerva:
Thank you for the clarification 👍

Issue #934, opened 2024-12-10
[Bug] ElasticSearch Connector doesn't respect limit or minRelevance parameters


basyonic:
### Context / Scenario

**Steps to reproduce:**
- Configure Kernel Memory ElasticSearch Connector
- Ingest document
- Use GetSimilarListAsync and pass limit and minRelevance Parameters

**Expected**
Limits and minRelevance are respected with max number of result match limit parameter and minRelevance Score match passed parameter

### What happened?

Limits and minRelevance are not respected.
API always return max of 15 results with any relevance score

### Importance

I cannot use Kernel Memory

### Platform, Language, Versions

Kernel Memory 0.94.241201.1

### Relevant log output

```shell

```

basyonic:
Bug fix in the below PR
https://github.com/microsoft/kernel-memory/pull/935

dluc:
@basyonic 

> API always return max of 15 results with any relevance score

which API do you mean here?

I left some comments in the PR, there are also bugs in the way the relevance score is handled.

PR #935, opened 2024-12-10
fix issue with elasticsearch extension to respect limit and minReleva…

basyonic:
…nce parameters when calling GetSimilarListAsync API

Fixing issue https://github.com/microsoft/kernel-memory/issues/934

using ElasticSearch Client .Size(limit) clause during query.
using .Similarity(min score) to fix minRelevance (after adjusting the score value based on cosine similarity as per https://www.elastic.co/guide/en/elasticsearch/reference/current/knn-search.html#knn-similarity-search)

### DIFF
--- a/extensions/Elasticsearch/Elasticsearch/ElasticsearchMemory.cs
+++ b/extensions/Elasticsearch/Elasticsearch/ElasticsearchMemory.cs
@@ -218,15 +218,20 @@ public class ElasticsearchMemory : IMemoryDb
         Embedding embedding = await this._embeddingGenerator.GenerateEmbeddingAsync(text, cancellationToken).ConfigureAwait(false);
         var coll = embedding.Data.ToArray();
 
+        //adjust min score for cosine similarity
+        //https://www.elastic.co/guide/en/elasticsearch/reference/current/knn-search.html#knn-similarity-search
+        float adjustedMinSimilarityScore = (float)(2 * minRelevance - 1);
+
         var resp = await this._client.SearchAsync<ElasticsearchMemoryRecord>(s =>
-                    s.Index(index)
+                    s.Index(index).Size(limit)
                         .Knn(qd =>
                         {
                             qd.k(limit)
                                 .Filter(q => this.ConvertTagFilters(q, filters))
                                 .NumCandidates(limit + 100)
                                 .Field(x => x.Vector)
-                                .QueryVector(coll);
+                                .QueryVector(coll)
+                                .Similarity(adjustedMinSimilarityScore);
                         }),
                 cancellationToken)
             .ConfigureAwait(false);


basyonic:

extensions/Elasticsearch/Elasticsearch/ElasticsearchMemory.cs
```diff
@@ -218,15 +218,20 @@ public async Task<string> UpsertAsync(
         Embedding embedding = await this._embeddingGenerator.GenerateEmbeddingAsync(text, cancellationToken).ConfigureAwait(false);
         var coll = embedding.Data.ToArray();
 
+        //adjust min score for cosine similarity
+        //https://www.elastic.co/guide/en/elasticsearch/reference/current/knn-search.html#knn-similarity-search
+        float adjustedMinSimilarityScore = (float)(2 * minRelevance - 1);
```
dluc, line 223: This formula transforms from score to cosine similarity, however internally ES works with score.

I believe this should be the inverse, e.g. `float adjustedMinSimilarityScore = (float)((minRelevance + 1) / 2)`

By the way, the formula `(2 * minRelevance - 1)` should be used in line 253, replacing `hit.Score` with the actual cosine similarity. The value returned is incorrect and doesn't match the similarity returned by other memory connectors.

extensions/Elasticsearch/Elasticsearch/ElasticsearchMemory.cs
```diff
@@ -218,15 +218,20 @@ public async Task<string> UpsertAsync(
...
+
         var resp = await this._client.SearchAsync<ElasticsearchMemoryRecord>(s =>
-                    s.Index(index)
+                    s.Index(index).Size(limit)
```
dluc, line 226: `qd.k(limit)` already sets a limit, isn't this redundant? should we remove the `limit` param in `qd.k(limit)`?

extensions/Elasticsearch/Elasticsearch.FunctionalTests/Additional/KernelMemoryTests.cs
```diff
@@ -25,6 +25,61 @@ public KernelMemoryTests(IConfiguration cfg, ITestOutputHelper output)
...
+
+        this.Output.WriteLine($"{partitions.Count}, Max Relevance: {maxRelevance}, Min Relevance: {minRelevance}");
+        Assert.True((partitions.Count > 5));
+        Assert.True((minRelevance < 0.82));
+
+        //limit 5:
+        results = await this.KernelMemory.SearchAsync("What is Silicon?", index: indexName, null, null, minRelevance: 0, limit: 5);
+        partitions = results.Results[0].Partitions;
+        maxRelevance = partitions.Max(p => p.Relevance);
+        minRelevance = partitions.Min(p => p.Relevance);
```
dluc, line 67: The 2 vars are not used. Missing some assertions?

extensions/Elasticsearch/Elasticsearch.FunctionalTests/Additional/KernelMemoryTests.cs
```diff
@@ -25,6 +25,61 @@ public KernelMemoryTests(IConfiguration cfg, ITestOutputHelper output)
...
+        Assert.True((partitions.Count == 5));
+
+        //relevance 0.82:
+        results = await this.KernelMemory.SearchAsync("What is Silicon?", index: indexName, null, null, minRelevance: 0.82, limit: 10);
+        partitions = results.Results[0].Partitions;
+        maxRelevance = partitions.Max(p => p.Relevance);
```
dluc, line 73: The var is not used. Missing an assertion?

extensions/Elasticsearch/Elasticsearch.FunctionalTests/Additional/TestsHelper.cs
```diff
@@ -18,6 +18,7 @@ internal static class TestsHelper
     public const string LoremIpsumFileName = "Data/file3-lorem-ipsum.docx";
     public const string NASANewsFileName = "data/file5-NASA-news.pdf";
     public const string SKReadmeFileName = "Data/file4-SK-Readme.pdf";
+    public const string SKWikipediaSciliconFileName = "Data/file7-Wikipedia-Silicon.txt";
```
dluc, line 21:
```suggestion
    public const string SKWikipediaSiliconFileName = "Data/file7-Wikipedia-Silicon.txt";
```

extensions/Elasticsearch/Elasticsearch.FunctionalTests/Additional/KernelMemoryTests.cs
```diff
@@ -25,6 +25,61 @@ public KernelMemoryTests(IConfiguration cfg, ITestOutputHelper output)
...
+    {
+        string indexName = nameof(this.ItSupportsLimitsAndMinRelevanceScoreAsync);
+        this.Output.WriteLine($"Index name: {indexName}");
+
+        const string Id = "ItSupportsLimitsAndMinRelevance-file7-Silicon-Carbon.txt";
+
+        this.Output.WriteLine("Uploading document");
+        await this.KernelMemory.ImportDocumentAsync(
+            new Document(Id)
+                .AddFile(TestsHelper.SKWikipediaSciliconFileName),
```
dluc, line 41:
```suggestion
                .AddFile(TestsHelper.SKWikipediaSiliconFileName),
```

extensions/Elasticsearch/Elasticsearch.FunctionalTests/Additional/KernelMemoryTests.cs
```diff
@@ -25,6 +25,61 @@ public KernelMemoryTests(IConfiguration cfg, ITestOutputHelper output)
 
     public IKernelMemory KernelMemory { get; }
 
+    [Fact]
+    [Trait("Category", "Elasticsearch")]
+    [SuppressMessage("Reliability", "CA2007:Consider calling ConfigureAwait on the awaited task", Justification = "<Pending>")]
```
dluc, line 30:
```suggestion
```
can be removed

dluc (changes requested):

extensions/Elasticsearch/Elasticsearch/ElasticsearchMemory.cs
basyonic: kNN search adds k matching documents to the search. So, if you set k=20, then it finds 20 matches. Then size takes the top scoring documents from these and returns them.
From my tests in real dataset. it was always returning 10 regardless of the k(limit) value. I think this is the default value of size parameter

extensions/Elasticsearch/Elasticsearch.FunctionalTests/Additional/KernelMemoryTests.cs
basyonic: here we are testing limit parameter, lines removed in updated PR

extensions/Elasticsearch/Elasticsearch.FunctionalTests/Additional/KernelMemoryTests.cs
basyonic: removed in updated PR

extensions/Elasticsearch/Elasticsearch/ElasticsearchMemory.cs
basyonic: As per https://www.elastic.co/guide/en/elasticsearch/reference/current/knn-search.html#knn-similarity-search
default similarity metric is cosine and I didn't see something defines another similarity metric in the plugin code.
From my tests I found this score adjustment for cosine returns correct results.
please let me know if I'm missing something here..
```A similarity value. This value determines the similarity metric used to score documents based on similarity between the query and document vector. For a list of available metrics, see the [similarity](https://www.elastic.co/guide/en/elasticsearch/reference/current/dense-vector.html#dense-vector-similarity) parameter documentation. **The similarity setting defaults to cosine**.```

dluc:
Pausing this while we wait for SK Vector Store relase and integration.

PR #940, opened 2024-12-12
Update README.md

TheovanKraay:
Azure Cosmos DB is not included in readme for supported semantic kernel vector stores (under extensions), but it is supported: https://learn.microsoft.com/en-us/semantic-kernel/concepts/vector-store-connectors/out-of-the-box-connectors/azure-cosmosdb-nosql-connector?pivots=programming-language-csharp

### DIFF
--- a/README.md
+++ b/README.md
@@ -348,7 +348,7 @@ storage, queues, vector stores, and LLMs to fit specific requirements.
 
 - **AI**: Azure OpenAI, OpenAI, ONNX, Ollama, Anthropic, Azure AI Document Intelligence, Azure AI
   Content Safety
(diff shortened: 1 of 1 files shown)


dluc:
Hi @TheovanKraay - Kernel Memory doesn’t use Semantic Kernel's memory/vector connectors because they lack advanced features like filtering. The connectors here are based on SK but add that functionality.

While SK supports various storage engines (SQLite, Chroma, Pinecone, Weaviate, etc.), without advanced filtering they can’t be used for Kernel Memory.

We could consider adding Cosmos DB support, but **my recommendation is to extend SK's vector store filtering capabilities** so we can avoid duplicate code and support more storage types right away.

PR #958, opened 2024-12-29
PosgreSQL hybrid search

SignalRT:
This PR Includes:
- Text search index
- Hybrid Search configuration
- Vector or hybrid search

Vector search do not provide the best results in an important number of scenarios. Hybrid search provides better results.

This change includes a new parameter for activate hybrid search on PostgreSQL extension. This parameters defaults to the previous implementation (vector search).
The search will use vector search or hybrid search depending on the parameter.

### DIFF
--- a/extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
+++ b/extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
@@ -59,6 +59,14 @@ internal sealed class PostgresDbClient : IDisposable, IAsyncDisposable
 
         this._columnsListNoEmbeddings = $"{this._colId},{this._colTags},{this._colContent},{this._colPayload}";
         this._columnsListWithEmbeddings = $"{this._colId},{this._colTags},{this._colContent},{this._colPayload},{this._colEmbedding}";
+        this._columnsListHybrid = $"{this._colId},{this._colTags},{this._colContent},{this._colPayload},{this._colEmbedding}";
+        this._columnsListHybridCoalesce = $@"
+                                            COALESCE(semantic_search.{this._colId}, keyword_search.{this._colId}) AS {this._colId},
+                                            COALESCE(semantic_search.{this._colTags}, keyword_search.{this._colTags}) AS {this._colTags},
+                                            COALESCE(semantic_search.{this._colContent}, keyword_search.{this._colContent}) AS {this._colContent},
+                                            COALESCE(semantic_search.{this._colPayload}, keyword_search.{this._colPayload}) AS {this._colPayload},
+                                            COALESCE(semantic_search.{this._colEmbedding}, keyword_search.{this._colEmbedding}) AS {this._colEmbedding},
+                                         ";
 
         this._createTableSql = string.Empty;
         if (config.CreateTableSql?.Count > 0)
@@ -388,6 +396,7 @@ internal sealed class PostgresDbClient : IDisposable, IAsyncDisposable
     /// Get a list of records
     /// </summary>
     /// <param name="tableName">Table containing the records to fetch</param>
+    /// <param name="query">Prompt query. Only used in the case of hybrid search</param>
     /// <param name="target">Source vector to compare for similarity</param>
     /// <param name="minSimilarity">Minimum similarity threshold</param>
     /// <param name="filterSql">SQL filter to apply</param>
@@ -395,9 +404,11 @@ internal sealed class PostgresDbClient : IDisposable, IAsyncDisposable
     /// <param name="limit">Max number of records to retrieve</param>
     /// <param name="offset">Records to skip from the top</param>
     /// <param name="withEmbeddings">Whether to include embedding vectors</param>
+    /// <param name="useHybridSearch">Whether to use hybrid search or vector search</param>
     /// <param name="cancellationToken">Async task cancellation token</param>
     public async IAsyncEnumerable<(PostgresMemoryRecord record, double similarity)> GetSimilarAsync(
         string tableName,
+        string query,
         Vector target,
         double minSimilarity,
         string? filterSql = null,
@@ -405,6 +416,7 @@ internal sealed class PostgresDbClient : IDisposable, IAsyncDisposable
         int limit = 1,
         int offset = 0,
         bool withEmbeddings = false,
+        bool useHybridSearch = false,
         [EnumeratorCancellation] CancellationToken cancellationToken = default)
     {
         tableName = this.WithSchemaAndTableNamePrefix(tableName);
@@ -413,6 +425,8 @@ internal sealed class PostgresDbClient : IDisposable, IAsyncDisposable
 
         // Column names
         string columns = withEmbeddings ? this._columnsListWithEmbeddings : this._columnsListNoEmbeddings;
+        string columsnHibrid = this._columnsListHybrid;
+        string columnsListHybridCoalesce = this._columnsListHybridCoalesce;
 
         // Filtering logic, including filter by similarity
         filterSql = filterSql?.Trim().Replace(PostgresSchema.PlaceholdersTags, this._colTags, StringComparison.Ordinal);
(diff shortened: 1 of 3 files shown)


SignalRT:

extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
```diff
@@ -175,7 +183,8 @@ public async Task CreateTableAsync(
                                 {this._colContent}   TEXT DEFAULT '' NOT NULL,
                                 {this._colPayload}   JSONB DEFAULT '{{}}'::JSONB NOT NULL
                             );
-                            CREATE INDEX IF NOT EXISTS idx_tags ON {tableName} USING GIN({this._colTags});
+                            CREATE INDEX IF NOT EXISTS {tableName}_idx_tags ON {tableName} USING GIN({this._colTags});
+                            CREATE INDEX IF NOT EXISTS {tableName}_idx_content ON {tableName} USING GIN(to_tsvector('english',{this._colContent}));
```
dluc, line 187: Should the language be configurable?

dluc, line 187: could you update this file too https://github.com/microsoft/kernel-memory/blob/main/extensions/Postgres/README.md please?

extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
```diff
@@ -413,6 +426,8 @@ DO UPDATE SET
 
         // Column names
         string columns = withEmbeddings ? this._columnsListWithEmbeddings : this._columnsListNoEmbeddings;
+        string columnsHibrid = this._columnsListHybrid;
```
dluc, line 429: nit: var name typo `columnsHibrid` => `columnsHybrid`

extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
```diff
@@ -440,16 +456,51 @@ DO UPDATE SET
...
+                            FROM {tableName}
+                            WHERE {filterSql}
+                            ORDER BY {this._colEmbedding} <=> @embedding
+                            LIMIT @limit
+                        ),
+                        keyword_search AS (
+                            SELECT {columnsHibrid}, RANK () OVER (ORDER BY ts_rank_cd(to_tsvector('english', {this._colContent}), query) DESC)
+                            FROM {tableName}, plainto_tsquery('english', @query) query
+                            WHERE  {filterSqlHybridText} AND to_tsvector('english', {this._colContent}) @@ query
+                            ORDER BY ts_rank_cd(to_tsvector('english', {this._colContent}), query) DESC
```
dluc, line 475: ditto: language should be configurable

extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
```diff
@@ -440,16 +456,51 @@ DO UPDATE SET
...
+                            LIMIT @limit
+                        )
+                        SELECT
+                            {columnsListHybridCoalesce}
+                            COALESCE(1.0 / (60 + semantic_search.rank), 0.0) +
+                            COALESCE(1.0 / (60 + keyword_search.rank), 0.0) AS {colDistance}
```
dluc, line 481: could you document or point to documentation explaining these formulas? e.g.  what is `60`?

extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
```diff
@@ -59,6 +59,14 @@ public PostgresDbClient(PostgresConfig config, ILoggerFactory? loggerFactory = n
 
         this._columnsListNoEmbeddings = $"{this._colId},{this._colTags},{this._colContent},{this._colPayload}";
         this._columnsListWithEmbeddings = $"{this._colId},{this._colTags},{this._colContent},{this._colPayload},{this._colEmbedding}";
+        this._columnsListHybrid = $"{this._colId},{this._colTags},{this._colContent},{this._colPayload},{this._colEmbedding}";
+        this._columnsListHybridCoalesce = $@"
+                                            COALESCE(semantic_search.{this._colId}, keyword_search.{this._colId}) AS {this._colId},
+                                            COALESCE(semantic_search.{this._colTags}, keyword_search.{this._colTags}) AS {this._colTags},
+                                            COALESCE(semantic_search.{this._colContent}, keyword_search.{this._colContent}) AS {this._colContent},
+                                            COALESCE(semantic_search.{this._colPayload}, keyword_search.{this._colPayload}) AS {this._colPayload},
+                                            COALESCE(semantic_search.{this._colEmbedding}, keyword_search.{this._colEmbedding}) AS {this._colEmbedding},
```
dluc, line 68: nit: remove the trailing comma (consistency with the way these values are used)

dluc (changes requested):
Looks almost ready to merge, only a few minor tweaks, please see the comments inline. Two most important:
- configurable language
- documenting the new SQL and the hard coded calculations

dluc:
Looks like the PR got stale, with some unsolved errors and comments. We might have to archive it unless someone can kindly complete the task.

SignalRT:
> Looks like the PR got stale, with some unsolved errors and comments. We might have to archive it unless someone can kindly complete the task.

I will resolve the comments this weekend.

extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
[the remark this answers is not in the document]
SignalRT: That's right the index should be build manually

extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
```diff
@@ -413,6 +426,8 @@ DO UPDATE SET
 
         // Column names
         string columns = withEmbeddings ? this._columnsListWithEmbeddings : this._columnsListNoEmbeddings;
+        string columnsHibrid = this._columnsListHybrid;
+        string columnsListHybridCoalesce = this._columnsListHybridCoalesce;
```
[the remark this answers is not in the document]
SignalRT: Removed

extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
SignalRT: I parametrized that values and I will update documentation

extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
SignalRT: Removed trailing comman

extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
SignalRT: Done

SignalRT (commented):
I finished to review the changes

extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
SignalRT: It's configurable now

extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
SignalRT: It's configurable now

extensions/Postgres/Postgres/PostgresConfig.cs
```diff
@@ -107,6 +107,23 @@ public class PostgresConfig
...
+    public bool UseHybridSearch { get; set; } = false;
+
+    /// <summary>
+    /// Defines the dictionary language the make the textual part of the hybrid Search in postgresql
+    /// see:
+    /// </summary>
+    public string TextSearchLanguage { get; set; } = "english";
+
+    /// <summary>
+    /// Reciprocal Ranked Fusion to score results of Hybrid Search
```
dluc, line 123: what is the "K" for? 

"RRF" => "Reciprocal Ranked Fusion"
"RRFK" => ?

pls add a link to the documentation

extensions/Postgres/Postgres/PostgresConfig.cs
```diff
@@ -107,6 +107,23 @@ public class PostgresConfig
...
 
+    /// <summary>
+    /// Important: when using hybrid search, relevance scores
+    /// are very different from when using just vector search.
+    /// </summary>
+    public bool UseHybridSearch { get; set; } = false;
+
+    /// <summary>
+    /// Defines the dictionary language the make the textual part of the hybrid Search in postgresql
+    /// see:
```
dluc, line 118: missing link

dluc:
TODO: 
1. when hybrid search is disabled (default) the code needs to be backward compatible, otherwise people using Postgres and upgrading KM will get a broken deployment. 
2. We need to define what happens if one enables hybrid search on an existing deployment, considering that KM will have to deal with tables with different schemas

extensions/Postgres/Postgres/PostgresConfig.cs
SignalRT: Done

SignalRT (commented):
Missed link and a query to check the installed dictionaries

aropb:
Suggestions for improvement:

-- Add a setting to specify normalization in ts_rank_cd():
https://www.postgresql.org/docs/current/textsearch-controls.html#TEXTSEARCH-RANKING

ts_rank_cd([ weights float4[], ] vector tsvector, query tsquery [, normalization integer ]) returns float4

normalization - this is an important parameter (however, then the calculation of the relevance value will change.).

0 (the default) ignores the document length
1 divides the rank by 1 + the logarithm of the document length
2 divides the rank by the document length
4 divides the rank by the mean harmonic distance between extents (this is implemented only by ts_rank_cd)
8 divides the rank by the number of unique words in document
16 divides the rank by 1 + the logarithm of the number of unique words in document
32 divides the rank by itself + 1

-- Add a setting to specify the Parsing query (websearch_to_tsquery, plainto_tsquery, phraseto_tsquery):
https://www.postgresql.org/docs/current/textsearch-controls.html#TEXTSEARCH-PARSING-QUERIES

-- I suggest using my example of creating a table:

  -- add a column for quick search: content_vector 
  -- add an index for embedding search: idx_embedding 

            BEGIN;
            SELECT pg_advisory_xact_lock(%%lock_id%%);

            CREATE TABLE IF NOT EXISTS %%table_name%% (
              id                TEXT NOT NULL PRIMARY KEY,
              tags              TEXT[] DEFAULT '{{}}'::TEXT[] NOT NULL,
              payload           JSONB DEFAULT '{{}}'::JSONB NOT NULL,
              content           TEXT DEFAULT '' NOT NULL,
              embedding         vector(%%vector_size%%),
              content_vector    tsvector GENERATED ALWAYS AS (to_tsvector('{LocaleConfig}', coalesce(content, ''))) STORED
            );

            CREATE INDEX IF NOT EXISTS idx_tags ON %%table_name%% USING gin (tags);
            CREATE INDEX IF NOT EXISTS idx_content_vector ON %%table_name%% USING gin (content_vector);
            CREATE INDEX IF NOT EXISTS idx_embedding ON %%table_name%% USING hnsw (embedding vector_cosine_ops);

            COMMIT;

**-- It is important that the new hybrid search does not interfere with the work without it. Including with a custom SQL script for creating a table!**

SignalRT:
> TODO:
> 
> 1. when hybrid search is disabled (default) the code needs to be backward compatible, otherwise people using Postgres and upgrading KM will get a broken deployment.
> 2. We need to define what happens if one enables hybrid search on an existing deployment, considering that KM will have to deal with tables with different schemas

@dluc ,

1.In my opinion the code is backward compatible.
2. I add a note in the documentation to explain how to create the index manually. This should also be done if you change the language of the index.

extensions/Postgres/Postgres/Internals/PostgresDbClient.cs
```diff
@@ -440,16 +461,51 @@ DO UPDATE SET
...
+                        cmd.CommandText = @$"
+                        WITH semantic_search AS (
+                            SELECT {this._columnsListHybrid}, RANK () OVER (ORDER BY {this._colEmbedding} <=> @embedding) AS rank
+                            FROM {tableName}
+                            WHERE {filterSql}
+                            ORDER BY {this._colEmbedding} <=> @embedding
+                            LIMIT @limit
+                        ),
+                        keyword_search AS (
+                            SELECT {this._columnsListHybrid}, RANK () OVER (ORDER BY ts_rank_cd(to_tsvector('english', {this._colContent}), query) DESC)
```
aropb, line 477: to_tsvector('english', {this._colContent} -- That's not very good! Each time a query is made, pgsql calculates a vector and the language specified here is clearly English, not a variable.

dluc:
Parking this for now, while we wait for SK Vector Store release. This feature should be available for free.

SignalRT:
@dluc , 

I was unable to find hybrid search support in **Semantic Kernel**. The only reference I came across is [this issue](https://github.com/microsoft/semantic-kernel/issues/11084), which includes the following comment:

image

I will move this PR to an Extensions Package to be able use it.

dluc:
@SignalRT yup, the work is still in progress, it hasn't started for PG but you can see how it will work checking the Azure AI Search connector:

https://github.com/microsoft/semantic-kernel/blob/e96c693851a2e4be14a9e7fd835c4087b063dc58/dotnet/src/Connectors/Connectors.Memory.AzureAISearch/AzureAISearchVectorStoreRecordCollection.cs#L440-L489

The work remaining for PG is implementing `IKeywordHybridSearch` in `PostgresVectorStoreRecordCollection`

https://github.com/microsoft/semantic-kernel/blob/e96c693851a2e4be14a9e7fd835c4087b063dc58/dotnet/src/Connectors/Connectors.Memory.Postgres/PostgresVectorStoreRecordCollection.cs#L20C14-L20C49

PR #959, opened 2025-01-02
fix: Resolve streaming bug in MemoryWebClient

westdavidr:
- Changed from PostAsync to SendAsync for correct request handling.
- Adjusted HttpCompletionOption for proper streaming behavior.

Resolves issue with previous streaming setup.

Changed from PostAsync to SendAsync for correct request handling and added the HttpCompletionOption so the operation completes as soon as the initial headers are sent.


dluc (approved):

dluc:
For the record, I'd consider this a performance improvement rather than a fix. E.g. the client was working with the previous implementation too.

PR #999, opened 2025-02-10
Upgrade dependencies

dluc:


dluc:
Looks like upgrading XUnit is a major task -- Reducing scope, see #1000

PR #1009, opened 2025-02-21
Install ICU in Docker image to support SQL Server memory

alexmg:

SQL Server memory cannot be configured when using the Docker image because it depends on ICU which is not installed in the `dotnet/aspnet:alpine` image. This results in a `System.NotSupportedException` being thrown at runtime when `SqlConnection` attempts to open a connection.

```
System.NotSupportedException: Globalization Invariant Mode is not supported.
    at Microsoft.Data.SqlClient.SqlConnection.TryOpen(TaskCompletionSource`1 retry, SqlConnectionOverrides overrides)
    at Microsoft.Data.SqlClient.SqlConnection.InternalOpenAsync(CancellationToken cancellationToken)
```

Update the `Dockerfile` to install `icu-libs` and set the `DOTNET_SYSTEM_GLOBALIZATION_INVARIANT` environment variable as outlined in the [Enabling (or disabling) globalization functionality](https://github.com/dotnet/dotnet-docker/blob/main/samples/enable-globalization.md) document.

This problem is discussed extensively in dotnet/SqlClient#220.

Building the Docker image was also failing in Docker Desktop on Windows because of the separator used in the `chown` command. This has been changed from `.` to `:` which should provide greater compatibility.

https://www.gnu.org/software/coreutils/manual/html_node/chown-invocation.html#chown-invocation

>  Some older scripts may still use `.` in place of the `:` separator. POSIX 1003.1-2001 (see [Standards conformance](https://www.gnu.org/software/coreutils/manual/html_node/Standards-conformance.html)) does not require support for that, but for backward compatibility GNU chown supports `.` so long as no ambiguity results, although it issues a warning and support may be removed in future versions. New scripts should avoid the use of `.` because it is not portable, and because it has undesirable results if the entire owner`.`group happens to identify a user whose name contains `.`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,6 +40,16 @@ RUN \
     # Allow user to access the build
     chown -R $USER.$USER /app
 
+# Use of Microsoft.Data.SqlClient requires ICU to be installed
+# https://github.com/dotnet/dotnet-docker/blob/main/samples/enable-globalization.md
+ENV \
+    DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=false \
+    LC_ALL=en_US.UTF-8 \
+    LANG=en_US.UTF-8
+RUN apk add --no-cache \
+    icu-data-full \
+    icu-libs
+
 COPY --from=build --chown=km:km --chmod=0550 /app/publish .
 
 #########################################################################
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,17 +38,12 @@ RUN \
     #Debian: useradd --create-home --user-group $USER --shell /bin/bash && \
     adduser -D -h /app -s /bin/sh $USER && \
     # Allow user to access the build
-    chown -R $USER.$USER /app
+    chown -R $USER:$USER /app
 
 # Use of Microsoft.Data.SqlClient requires ICU to be installed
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -38,7 +38,12 @@ RUN \
     #Debian: useradd --create-home --user-group $USER --shell /bin/bash && \
     adduser -D -h /app -s /bin/sh $USER && \
     # Allow user to access the build
-    chown -R $USER.$USER /app
+    chown -R $USER:$USER /app
+
+# Use of Microsoft.Data.SqlClient requires ICU to be installed
+# https://github.com/dotnet/dotnet-docker/blob/main/samples/enable-globalization.md
+ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=false
+RUN apk add --no-cache icu-libs
```
dluc, line 46: Could you move the env down after the other envs?
Also, merge the apk command with the RUN above? (to avoid extra layers)

alexmg: Changes have been pushed. I actually started with the command in the existing RUN but moved them out to keep the two settings together with the link. I've added a relevant comment to each line instead.

dluc (approved):

dluc:
Note: the resulting img is ~7MB bigger.

PR #1032, opened 2025-03-17
Add OpenAPI specification for /upload endpoint.

FabianSchurig:
Hi @dluc

Currently the specification of the `/upload` endpoint is not automatically generated.

My change manually adds the specification to `service/Service.AspNetCore/WebAPIEndpoints.cs` which enables the SwaggerUI to provide a _Try it out_ option with the upload feature and setting the request body.

> [!IMPORTANT]
>  Furthermore, I will need the generated change of the `swagger.json` to automatically generate the python client with `openapi-python-generator` in a follow up PR. The client will be auto generated based on the OpenAPI spec.

Here an example of how it is looks like and works:
image


FabianSchurig:

dluc (approved):

dluc:
Big Thanks @FabianSchurig 👍

PR #1039, opened 2025-03-24
Add ForcePathStyle support to AWSS3Config for MinIO compatibility

eersinyildizz:
This PR adds support for `ForcePathStyle` configuration in `AWSS3Config`

- Added a `ForcePathStyle` boolean property to `AWSS3Config` (default: `false`)
- Updated `AWSS3Storage` to pass the value to the `AmazonS3Config.ForcePathStyle` property
- Maintains backward compatibility with AWS S3 by leaving the default behavior unchanged
- Improves compatibility with services like MinIO, LocalStack, and others that require path-style requests

This change is fully backward compatible and has no effect on current configurations unless explicitly enabled.

### DIFF
--- a/extensions/AWS/S3/AWSS3Config.cs
+++ b/extensions/AWS/S3/AWSS3Config.cs
@@ -39,6 +39,12 @@ public class AWSS3Config
     /// </summary>
     public string BucketName { get; set; } = string.Empty;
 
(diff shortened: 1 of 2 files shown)


eersinyildizz:

dluc (approved):

PR #1063, opened 2025-05-08
Added fallback for MAXSEARCHRESULTS in RedisMemory:GetListAsync

AnnoDeJong:
Azure Redis Enterprise does not support FT.CONFIG commands which causes the method to fail completely and as this is used when deleting a document it prevents a document from being deleted.

Added a fallback value as well as moved the configuration task into the if statement.


AnnoDeJong:
> * Submissions in the course of work for my

raiderss:
Impressive project! Eyes Store community would find this quite useful for FiveM development.

Issue #677, opened 2024-06-25
Are `.doc` files supported?


gmantri:
### Context / Scenario

We have some Microsoft Word documents that are in old format (`.doc`). What we are seeing is that when try to use those documents, Kernel Memory fails to answer the questions from those documents. When we convert those documents to `.docx` format, everything works great.

[microsoft.docx](https://github.com/user-attachments/files/15979329/microsoft.docx)

### What happened?

Our expectation was that both `.doc` and `.docx` files should work but that is not happening. `.doc` files do not work but `.docx` file work.

### Importance

a fix would make my life easier

### Platform, Language, Versions

Microsoft.KernelMemory.Core - 0.61.240524.1
Microsoft.SemanticKernel - 1.15.0

dluc:
Hi @gmantri the old .doc format is not supported sorry. Aside from converting files manually, you could:

* option 1: add a custom handler in the ingestion pipeline to convert .doc files to PDF/DOCX, before extract text step. See `steps` parameter in the Import API.
* option 2: add a custom decoder for .doc files, doing the same, e.g. converting to PDF/DOCX and returning the test. There should be some examples in the `examples` folder about adding custom decoders.

gmantri:
@dluc - Thanks. Is there a list of file types supported by Kernel Memory. All I could find was this: https://github.com/microsoft/kernel-memory?tab=readme-ov-file#kernel-memory-km-and-sk-semantic-memory-sm and it only talks about the file types at a high level (e.g. Word instead of .docx and not .doc). Having this list will be really helpful.

dluc:
The default list can be extrapolated from here https://github.com/microsoft/kernel-memory/blob/3d34260ae513af48030da9a56aa50b8e0162c6f8/service/Core/DataFormats/DependencyInjection.cs#L81

```csharp
        services.AddSingleton<IContentDecoder, TextDecoder>();
        services.AddSingleton<IContentDecoder, MarkDownDecoder>();
        services.AddSingleton<IContentDecoder, HtmlDecoder>();
        services.AddSingleton<IContentDecoder, PdfDecoder>();
        services.AddSingleton<IContentDecoder, ImageDecoder>();
        services.AddSingleton<IContentDecoder, MsExcelDecoder>();
        services.AddSingleton<IContentDecoder, MsPowerPointDecoder>();
        services.AddSingleton<IContentDecoder, MsWordDecoder>();
```

[...]

PR #1074, opened 2025-06-25
Adding support for legacy Word .doc files

luismanez:
This should fix issue #677 
Although .doc files are legacy, many organizations still have a lot of their knowledge base in .doc files. It'd be great to support that in kernel memory, helping to have that knowledge in a chatGPT solution.

This feature is using the existing open source library ```NPOI``` (https://github.com/nissl-lab/npoi), which seems is the best solution to deal with legacy Office documents (not only Word, but also Excel files). I use the library to extract all the text from the .doc file as plain text.

I've also created a Test for the new Decoder, and updated the existing 002-dotnet-Serverless sample to Index a legacy .doc file and retrieve an answer from it.

__Note__: to support some old Encodings (like Windows-1255), I had to add the package ```System.Text.Encoding.CodePages``` and load the encoding in the Decoder constructor.


dluc:

PR #1082, opened 2025-09-04
Limit GetSimilarRecordsList by index.

crwells:
For SQL Server 2025, the PrepareGetSimilarRecordsListQuery() method produces a query that does not account for the index that is being searched.  This only happens when NativeVectorSearch is true and the VectorQueryProvider is used.  The DefaultQueryProvider does not suffer from the same problem (though may benefit from the same fix performance-wise)  because it links to tables that are index-specific, like [{embeddingstable}_{index}] which naturally limits the memories via the inner join.

Simply add in the [collection]=@index criteria.


dluc:

Issue #1084, opened 2025-09-09
[Bug] Need to await calls to DecodeAsync(stream) from DecodeAsync(string filename) methods in most IContentDecoders


crwells:
### Context / Scenario

Two of the DecodeAsync methods are implemented like this on several of the IContentDecoder classes:

```
public Task<FileContent> DecodeAsync(string filename, CancellationToken cancellationToken = default)
{
    using var stream = File.OpenRead(filename);
    return this.DecodeAsync(stream, cancellationToken);
}

public Task<FileContent> DecodeAsync(BinaryData data, CancellationToken cancellationToken = default)
{
    using var stream = data.ToStream();
    return this.DecodeAsync(stream, cancellationToken);
}
```

Since these aren't async methods, and we aren't awaiting the inner call to this.DecodeAsync(stream, ct), the stream is disposed of before it is used (ObjectDisposedException in MarkDownDecoder.cs, on the call to reader.ReadToEndAsync, as an example).

### What happened?

Basic Reproduction:
```
var decoder = _decoders.LastOrDefault(d => d.SupportsMimeType(mimeType))
if (decoder != null)
{
        var content = await decoder.DecodeAsync(filePath).ConfigureAwait(false);
        return content.Sections;
}
```

This exception is received:
```
System.ObjectDisposedException: 'Cannot access a closed file.'
This exception was originally thrown at this call stack:
    System.ThrowHelper.ThrowObjectDisposedException_FileClosed()
    System.IO.FileStream.ReadAsync(System.Memory<byte>, System.Threading.CancellationToken)
    System.IO.StreamReader.ReadBufferAsync(System.Threading.CancellationToken)
    System.IO.StreamReader.ReadToEndAsyncInternal(System.Threading.CancellationToken)
    Microsoft.KernelMemory.DataFormats.Text.MarkDownDecoder.DecodeAsync(System.IO.Stream, System.Threading.CancellationToken)
   
``` 

Workaround:

```
var decoder = _decoders.LastOrDefault(d => d.SupportsMimeType(mimeType))
if (decoder != null)
{
    using (var stream = File.OpenRead(doc.FilePath))
    {
        var content = await decoder.DecodeAsync(stream).ConfigureAwait(false);
        return content.Sections;
    }
}
```

### Importance

a fix would make my life easier

### Platform, Language, Versions

C#
Microsoft.KernelMemory 0.98.250508.3

### Relevant log output

```shell

```

PR #1085, opened 2025-09-09
Added await when calling async DecodeAsync method with disposable item.

crwells:
Motivation and Context (Why the change? What's the scenario?)

Fix for bug #1084  
Need to await the async DecodeAsync method calls so that the stream object isn't disposed of before it is used.

Just need to await.


dluc:

PR #1114, opened 2025-12-18
Complete feature: embeddings batching, retries, managed identity

dluc:
Complete Embedding Generators and Cache feature by implementing remaining Must-Have acceptance criteria.

Changes:
- Add configurable embeddings batch sizing (default 10) via `EmbeddingsConfig.batchSize` and chunking for batch-capable providers.
- Add shared transient HTTP retry/backoff (honors `Retry-After`) and use it in OpenAI/Azure/HF/Ollama providers.
- Add Azure OpenAI managed identity auth (`DefaultAzureCredential`) alongside API key auth.
- Add HuggingFace `HF_TOKEN` environment variable fallback.
- Add/adjust unit + integration tests and stabilize env-var dependent tests.

### DIFF
--- a/src/Core/Config/AppConfig.cs
+++ b/src/Core/Config/AppConfig.cs
@@ -10,7 +10,9 @@ namespace KernelMemory.Core.Config;
 /// Root configuration for Kernel Memory application
 /// Loaded from ~/.km/config.json or custom path
 /// </summary>
+#pragma warning disable CA1724 // Conflicts with Microsoft.Identity.Client.AppConfig when Azure.Identity is referenced.
 public sealed class AppConfig : IValidatable
+#pragma warning restore CA1724
 {
     /// <summary>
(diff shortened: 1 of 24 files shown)


dluc:
Restored  to upstream behavior (reverted the FTS-only config + removed the accidental change to Ollama gating).

dluc:
Restored `tests/Main.Tests/Integration/SearchProcessTests.cs` to upstream behavior (reverted the FTS-only config change and restored the original Ollama/`OLLAMA_AVAILABLE` gating).

dluc:
Pushed a fix for the retry regression: `HttpRetryPolicy` now fails fast on connection-refused/host-not-found/network-unreachable and enforces per-attempt timeouts (Ollama uses a low timeout).

This prevents `coverage.sh`/test runs from appearing hung when Ollama is stopped.

